PR #3, opened 2021-05-18
build: remove the framework base version from dynamic versions

pvlugter:
When determining the SDK version, no longer use the Akka Serverless framework base version. Use regular dynamic versioning.


ennru (approved):
LGTM.

PR #8, opened 2021-05-19
docs: setup docs deploy with api docs and attributes

pvlugter:
Use the same approach as we did for Cloudstate, publishing doc sources to a branch, including dynamically generated / managed sources. Currently just the API docs and version attributes. Deleted the old doc content and examples.

Scripts based on the [cloudstate-antora](https://github.com/cloudstateio/cloudstate-antora) scripts. Docs get published to [`docs/current`](https://github.com/lightbend/akkaserverless-javascript-sdk/tree/docs/current) branch. Docs version is always the nearest version tag (so that we can easily publish doc fixes as necessary manually). Pushed the initial version manually already, to use in the Akka Serverless docs.


ennru (approved):
LGTM.

PR #12, opened 2021-05-23
doc: start copying doc examples directly from sample apps

pvlugter:
The `js-customer-registry` sample app is already annotated with doc tags, and has been manually copied into the docs repo. Copy the sources for this sample into the doc examples automatically instead.

Already published this manually to the `docs/current` branch.


ennru (approved):
LGTM.

PR #17, opened 2021-06-03
fix: deserialize protobuf any with undefined values (no bytes)

pvlugter:
Noticed when sending requests with `grpcurl` and empty body. These arrive as protobuf any with undefined values, which would crash the service. Default to an empty array/buffer and check that these are deserialised as an empty message.


octonato (approved):
LGTM

PR #20, opened 2021-06-07
Let action return the metadata class, instead of raw metadata

jroper:


sdk/src/action-support.js
```diff
@@ -86,7 +86,7 @@ class ActionHandler {
         return call.cancelled;
       },
       get metadata() {
-        return metadata;
+        return metadataObject;
```
jroper, line 89: On line 79-81 we create this object for this purpose, and should return it here.

ennru (approved):
LGTM.

PR #22, opened 2021-06-07
feat: basic type definition generation from jsdoc

janory:
Fixes this issue https://github.com/lightbend/akkaserverless-framework/issues/689.

In order to make the JSDoc -> .d.ts generation work, I had to change a few things:
- fix JSDoc annotations/comments where the generator yielded an error (please verify if the changes make sense)
- use a patched tsd-jsdoc generator because of this issue: https://github.com/englercj/tsd-jsdoc/issues/89
- [fork and create a package](https://github.com/janory/tsd-jsdoc/packages/833291) from [englercj's PR](https://github.com/englercj/tsd-jsdoc/pull/144/files#diff-9af2ecc9ea9c57456a81fadd6bf93a6d974c4aa1421e8bf912ce080d70be4fcb)
- even the forked version didn't work fully so had to [monkey patch it](https://github.com/janory/tsd-jsdoc/commit/d5ebc54eb6e624e0bc7e56a8f73f27296a2dada3) before publishing
- add the .d.ts generation to the `prepare.sh` script with some post processing (mostly regex search & replaces) to get the correct types

Hint: I've forked the tsd-jsdoc library under my account instead of the LB organisation, because I was not sure about the licensing (the original license is MIT, so it should fine though) and legal implications. It's already a fork of a fork and I wanted to make this work quickly, I thinks it's fine to have it under my account, because it's not too much different as using package from foreign people from NPM, but I can transfer the ownership of the repo to LB if we want to.

This is the first and very basic version of the type definition support, there are plenty of TODOs which we can do to improve the generated types. 

I see the following short-term TODOs which could help us to get better typing:
- there are still lots of warnings by JSDoc during the type definition generation, we should analyze and fix those, because the generator often just falls back to `any`
- try to eliminate as much `any`s as possible by locally solving the [{object} -> any issue](https://github.com/microsoft/TypeScript/issues/18396)
- remove the dirty regexes and solve those issues on the `tsd-jsdoc` library level

A long-term TODO would be to transform the SDK's codebase to Typescript and don't rely on the JSDoc for type generation.
Since we already generate fairly accurate types based on the JSDoc this transition could happen easier and of course gradually. Based on the decision of how and when do we want to make this step, we should consider to put as less effort into the short-term TODOs as possible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+.idea
--- a/sdk/.gitignore
+++ b/sdk/.gitignore
@@ -2,3 +2,4 @@ proto
 protoc
 apidocs
 user-function.desc
+index.d.ts
--- a/sdk/.npmrc
+++ b/sdk/.npmrc
@@ -1,1 +1,2 @@
 engine-strict=true
+@janory:registry=https://npm.pkg.github.com
--- a/sdk/.npmrc
+++ b/sdk/.npmrc
@@ -1,2 +1,2 @@
 engine-strict=true
-@janory:registry=https://npm.pkg.github.com
+@lightbend:registry=https://npm.cloudsmith.io/lightbend/akkaserverless
--- a/sdk/bin/prepare.sh
+++ b/sdk/bin/prepare.sh
@@ -36,7 +36,7 @@ pbjs -t static-module -p ./proto -p ./protoc/include \
   | pbts -o ./proto/protobuf-bundle.d.ts -
 
 # Generate TS type definitions based on the JSDocs
-jsdoc -t ./node_modules/@janory/tsd-jsdoc/dist -c ./jsdoc.json -d .
+jsdoc -t ./node_modules/@lightbend/tsd-jsdoc/dist -c ./jsdoc.json -d .
 mv types.d.ts index.d.ts
 # There replacements are quite dirty, but even the patched tsd-jsdoc generator can't deal with these (mostly module related) issues currently
 perl -i -pe 's/declare module \"akkaserverless\"/declare module \"\@lightbend\/akkaserverless-javascript-sdk\"/g' index.d.ts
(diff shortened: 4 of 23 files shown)


janory:
cc @RoryStokes

janory:
CircleCI is currently failing, but I am already working on a solution to make this work (either publish somewhere else or add a token). In the meanwhile if you would like to try this out locally, please add a GH token which has the `read:packages` permission to your `~/.npmrc`.

pvlugter (commented):
LGTM. Agree on focusing more on the long-term transition to TypeScript, over putting much energy into the short-term fixes.

I can add a token for GitHub packages. But a token will also be required by anyone building this now, and also affects the docker builds for the TCK and samples. Shall we publish the forked `tsd-jsdoc` somewhere else (that doesn't require any auth)? We could use a Cloudsmith repository (where other Lightbend artifacts have been migrated to), rather than the main npm registry.

janory:
> We could use a Cloudsmith repository (where other Lightbend artifacts have been migrated to), rather than the main npm registry.

Done! 👍 
I've moved the package [to the LB org](https://github.com/lightbend/tsd-jsdoc) and [published to our Cloudsmith org](https://cloudsmith.io/~lightbend/repos/akkaserverless/packages/detail/npm/@lightbend%252Ftsd-jsdoc/2.5.1/).

.gitignore
```diff
@@ -1 +1,2 @@
 node_modules
+.idea
```
pvlugter, line 2: Not important, but good to add any IDE or editor ignores to your own global gitignore. People use so many different tools, and it saves putting them all into the project gitignore.

pvlugter (approved):

pvlugter:
> I've moved the repo [to the LB org](https://github.com/lightbend/tsd-jsdoc) and [published the package to our Cloudsmith org](https://cloudsmith.io/~lightbend/repos/akkaserverless/packages/detail/npm/@lightbend%252Ftsd-jsdoc/2.5.1/).

Cool 👍🏼

PR #24, opened 2021-06-08
Handle Async Commands

andreaTP:
There are still 4 tck tests failing I should look into.

I open this PR as a draft since I'm not satisfied with the design of this feature, still it was the best approach I have found without a major re-write.

I'm super open to discussions, feedback and suggestions!

### DIFF
--- a/sdk/package-lock.json
+++ b/sdk/package-lock.json
@@ -1,8 +1,6763 @@
 {
   "name": "@lightbend/akkaserverless-javascript-sdk",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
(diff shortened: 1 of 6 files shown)


pvlugter:
I feel like we should be able to support async handlers in a simpler way, with the changes focused on the command helper code. If the handler returns a promise ("thenable") then run the reply logic async on the promised value instead (including deactivating the context). But I haven't looked at it more closely before.

pvlugter:
After a discussion, it seems that refactoring things further to have a clearer structure, consolidating where the context is mutated and calls written and having the internals more clearly typed (such as returning promises of replies), will be useful both for making the async handling possible, as well as reading and maintaining the codebase going forward. Thanks @andreaTP 👍🏼

andreaTP:
Finalized in #43

Issue #7, opened 2021-05-19
Test the npm package


pvlugter:
So we can ensure that published packages have everything and work well, it would be useful to have integration tests run against a locally created package, rather than against a local directory. This could be done by packaging with `npm pack` then installing the tar directly, or un-tarring the package to a directory and then linking that directory for testing.

pvlugter:
I'll make this update, replacing the `npm link` that we now use, with `npm pack` + `npm install ...tgz`

PR #25, opened 2021-06-08
ci: remove sudo commands from circleci builds

pvlugter:
Remove the sudo commands from CI builds, which are likely causing permission issues.

Prefer [`npm ci`](https://docs.npmjs.com/cli/v7/commands/npm-ci) for the SDK. Use `nvm` for node version management.

I've left these linked SDK tests as standalone, but we could move the `npm link` up to the sdk build and test command at some point.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,8 @@ jobs:
             pushd samples/js-eventsourced-shopping-cart && npm ci && npm test && popd
             pushd samples/js-valueentity-shopping-cart && npm ci && npm test && popd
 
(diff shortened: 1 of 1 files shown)


pvlugter:
Okay, so the integration tests are fine, since we use a machine instance and `nvm`. The `cimg/node` docker image is using a globally installed node. Can switch to the user local prefix approach to avoid `sudo`.

We may also want to `pack` the SDK instead for fully testing, as in #7

octonato (approved):
LGTM

PR #31, opened 2021-06-15
chore: rename docs module from js-services to javascript

pvlugter:
Rename from `js-services` to `javascript`. Keep a copy at the old location during the transition. Already published to the [`docs/current` branch](https://github.com/lightbend/akkaserverless-javascript-sdk/tree/docs/current/docs/build/src/managed/modules).


ennru (approved):
LGTM.

PR #19, opened 2021-06-04
Value Entity example for the docs

andreaTP:
Following the example for Java developed here:

https://github.com/lightbend/akkaserverless-java-sdk/pull/11

This is the translation for JS.

PR #33, opened 2021-06-15
perf/fix: selective header forwarding to user function 

johanandren:
List of headers to be forwarded in component options.

Manually verified against the updated proxy.

This is waiting on a release of proxy with lightbend/akkaserverless-framework#729 merged.

### DIFF
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -86,7 +86,7 @@
     "print-version": "echo $npm_package_version",
     "watch-jsdoc": "npm-watch jsdoc",
     "jsdoc": "jsdoc -c jsdoc.json",
-    "test": "mocha ./test/**/*.* --recursive --unhandled-rejections=strict",
+    "test": "mocha ./test/*.test.* ./test/**/*.* --recursive --unhandled-rejections=strict",
     "integration-test": "npm run compile && mocha integration-test",
     "preintegration-test": "bin/compile-descriptor.js test/example.proto --descriptor_set_out=integration-test/user-function.desc",
     "prepare": "bin/prepare.sh",
--- a/sdk/src/action.js
+++ b/sdk/src/action.js
@@ -28,6 +28,7 @@ const actionServices = new ActionSupport();
  *
  * @typedef module:akkaserverless.Action~options
  * @property {array<string>} [includeDirs=["."]] The directories to include when looking up imported protobuf files.
+ * @property {array<string>} forwardHeaders request headers to be forwarded as metadata to the action
  */
 
 /**
--- a/sdk/src/akkaserverless.ts
+++ b/sdk/src/akkaserverless.ts
@@ -60,16 +60,22 @@ export interface EntityPassivationStrategy {
 }
 
 export interface ComponentOptions {
-  entityType: string;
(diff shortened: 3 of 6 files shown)


sdk/package-lock.json
```diff
@@ -1,7093 +1,8 @@
 {
   "name": "@lightbend/akkaserverless-javascript-sdk",
   "version": "0.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
pvlugter, line 4: The lock file was updated to v2 for npm 7 (in #19). We probably don't want to keep going back and forth, so using node 14 with upgraded npm 7 will be required locally now.

johanandren: Ah, accidentally committed mostly

andreaTP:
Not a big deal but I added some tests in #38 that can cover this use case:
https://github.com/lightbend/akkaserverless-javascript-sdk/pull/38/files#diff-a41d11ba5add430fe99954315fe0cf537270d4facfc493edf1abb84537af6ba6R144-R167

johanandren:
Note that this still is waiting for the next proxy release with the protocol change.

sdk/test/akkaserverless-test.js
```diff
@@ -15,6 +15,7 @@
  */
 
 const should = require('chai').should();
+const expect = require('chai').expect;
```
andreaTP, line 18: should we converge to one testing style?
I am not strongly opinionated but I do prefer to stitch with one "per-cdebase"

sdk/package-lock.json
andreaTP: still on `v1`

sdk/test/akkaserverless-test.js
```diff
@@ -154,6 +156,9 @@ At package.test.json:2:4:
     comp.getComponentType().should.equal('my-type');
     comp.getEntity().getEntityType().should.equal('my-entity-type');
     should.equal(comp.getEntity().getPassivationStrategy(), undefined);
+    expect(comp.getEntity().getForwardHeadersList()).to.have.same.members([
```
andreaTP, line 159: should we have a quick test for the two possible code-paths?

sdk/test/akkaserverless-test.js
johanandren: Sure, no strong preference, just didn't find API to express the same assertion with `should`, do you know how that would look for the this array should have the same elements as that array?

pvlugter: Should be able to just rewrite in `should` format:

```js
comp.getEntity().getForwardHeadersList().should.have.same.members([
  'x-my-header'
]);
```

sdk/test/akkaserverless-test.js
andreaTP: yup, is it supposed to happen at all?

johanandren: Yeah, action should be able to pick up headers as well

johanandren: Covered

pvlugter (approved):
LGTM

PR #34, opened 2021-06-15
Fix testkit MockEventSourcedEntity comment casing

beritou:


pvlugter (approved):

pvlugter:
This is from the code generation, so will also need to be updated in the template there.

PR #40, opened 2021-06-16
Include a process-uuid for quicker dev mode refresh/reconnect

johanandren:
SDK side of https://github.com/lightbend/akkaserverless-framework/pull/741


johanandren:
Closing this as we need something else that covers both dev and prod.

andreaTP:
Not sure if this fits the use case, but let me drop here a line, getting an ID from a File follows standard Linux practices.
E.g. in dev you can point to: `/dev/random` and in prod you can point to a static file (mounted or not from configmap / secrets). @johanandren

PR #41, opened 2021-06-16
get rid of allComponents

andreaTP:
Here we stop passing around `allComponents` to each and every class.

The only reason (at best of my understanding) we have been doing so was to look up the function in `effect-serializer`, but the method itself is passed as a first argument.

Anything I'm missing?

### DIFF
--- a/sdk/src/effect-serializer.js
+++ b/sdk/src/effect-serializer.js
@@ -17,58 +17,57 @@
 const AnySupport = require("./protobuf-any");
 const util = require("util");
 
-module.exports = class EffectSerializer {
+function serializeEffect(method, message, metadata) {
+  let serviceName, commandName;
+  // We support either the grpc method, or a protobufjs method being passed
+  if (typeof method.path === "string") {
+    const r = new RegExp("^/([^/]+)/([^/]+)$").exec(method.path);
+    if (r == null) {
+      throw new Error(util.format("Not a valid gRPC method path '%s' on object '%o'", method.path, method));
+    }
+    serviceName = r[1];
+    commandName = r[2];
+  } else if (method.type === "rpc") {
+    serviceName = fullName(method.parent);
+    commandName = method.name;
+  }
 
-  constructor() {}
+  const command = method;
+  if (command !== undefined) {
+    const payload = AnySupport.serialize(command.resolvedRequestType.create(message), false, false);
+    const effect = {
+      serviceName: serviceName,
+      commandName: commandName,
+      payload: payload
+    };
 
-  serializeEffect(method, message, metadata) {
-    let serviceName, commandName;
-    // We support either the grpc method, or a protobufjs method being passed
-    if (typeof method.path === "string") {
-      const r = new RegExp("^/([^/]+)/([^/]+)$").exec(method.path);
-      if (r == null) {
-        throw new Error(util.format("Not a valid gRPC method path '%s' on object '%o'", method.path, method));
+    if (metadata && metadata.entries) {
+      effect.metadata = {
+        entries: metadata.entries
       }
-      serviceName = r[1];
-      commandName = r[2];
-    } else if (method.type === "rpc") {
-      serviceName = this.fullName(method.parent);
-      commandName = method.name;
     }
 
-    const command = method;
-    if (command !== undefined) {
-      const payload = AnySupport.serialize(command.resolvedRequestType.create(message), false, false);
-      const effect = {
-        serviceName: serviceName,
-        commandName: commandName,
-        payload: payload
-      };
-
-      if (metadata && metadata.entries) {
-        effect.metadata = {
-          entries: metadata.entries
-        }
-      }
-
-      return effect;
-    } else {
-      throw new Error(util.format("Command [%s] unknown on service [%s].", commandName, serviceName))
-    }
+    return effect;
+  } else {
+    throw new Error(util.format("Command [%s] unknown on service [%s].", commandName, serviceName))
   }
+}
 
-  fullName(item) {
-    if (item.parent && item.parent.name !== "") {
-      return this.fullName(item.parent) + "." + item.name;
-    } else {
-      return item.name;
-    }
+function fullName(item) {
+  if (item.parent && item.parent.name !== "") {
+    return fullName(item.parent) + "." + item.name;
+  } else {
+    return item.name;
   }
+}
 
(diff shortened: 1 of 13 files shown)


sdk/src/effect-serializer.js
```diff
@@ -19,9 +19,7 @@ const util = require("util");
 
 module.exports = class EffectSerializer {
 
-  constructor(allComponents) {
-    this.allComponents = allComponents;
-  }
+  constructor() {}
```
johanandren, line 22: Is it still needed if it has no parameters and doesn't do anything? Also, since it doesn't have any state anymore it could be a singleton rather than something you create an instance of when you want to use it?

sdk/src/effect-serializer.js
```diff
@@ -38,30 +36,24 @@ module.exports = class EffectSerializer {
       commandName = method.name;
     }
```
johanandren, line 37: Since method can be either of these two types, does it really work without the component lookup?

I see all tests are passing but I wonder if they really exercise both types of method here.

sdk/src/effect-serializer.js
andreaTP: agree done 👍

sdk/src/effect-serializer.js
andreaTP: Good point! Do you know anyone able to answer this question?

In absence of further information, in between:
 - losing a feature that is not documented nor tested (aka. possibly introducing a bug)
 - removing technical debt by avoiding sharing everything with everyone

I would still pick the latter.

johanandren: Perhaps @pvlugter knows the rationale

johanandren: The test coverage here is definitely on the low end so you cannot count on anything not tested not being important. Not knowing what purpose something has is not the same as that thing being technical debt.

andreaTP: @johanandren agreed and that's the reason I stated:

> In absence of further information

Willing to receive more explanations here! 🙂 

On:

> being technical debt

I can say that passing around `allComponents` everywhere looks, to me, a shortcut (possibly due to pressure/lack of time) instead of a deliberate decision.

pvlugter: I haven't looked at this before, but it certainly looks like validating that the component service is registered is intentional here, and useful for both types (as strings or actual method objects). Adding @jroper.

But this should also be checked when the forward or side effect goes to the proxy (and hopefully the error is already nicely reported back to the user function too), in which case we could see the check here as unnecessary. Probably worthwhile trying this out to make sure that responding with a forward or side effect to an unknown service is obvious and well reported. The current approach would fail the command handling, while having the check proxy-side would have the command successfully handled and then an error reported.

Agree that the test coverage is thin for this SDK. We've been looking to add tests whenever working in an area.

jroper: It's been a long time, but I think the idea was that conceptually, we need a protobuf method descriptor. What we don't need is a fully resolved, grpc backend specific method. We shouldn't require users to select a gRPC library to use to do this, since you're not making a grpc call. Sure, we do use node-grpc underneath but that's an implementation detail, not something that we should expose to users. There are other libraries that we may decide to swap in/out in future for whatever reason (including perhaps the new js-only, no c++ code, library).

But protobufjs seems to be the de-facto standard for working with protobuf, so that's what we want to support, at a minimum. This is what it looks like to lookup a service with protobufjs (I think):

```javascript
const root = require("protobufjs").loadSync("myprotobuf.proto");
const method = root.lookupService("example.MyService").methods.MyMethod;
```

That will break with this change, since that method does not have a property called `resolvedRequestType`, it just references the request type by name.

The reason our TCK still passes is that our TCK uses the node-grpc objects, it actually uses an undocumented property on the entity called `service` which contains the gRPC service descriptor. Perhaps we should change that.

andreaTP: Can someone help me write a test to verify this feature? Is it documented or used anywhere?

@jroper how would it be this feature design without depending on `protobufjs` and without having to pass around `allComponents`?

jroper: @andreaTP Not sure what you're talking about. Why would we aim to not pass around `allComponents`? Is passing `allComponents` around a problem? Unless we make users serialise the messages passed in effects/forwards themselves, we need a way of looking up the serialisers for the effects. The serialisers are part of the gRPC services that each component exposes. We could extract the services from `allComponents` and instead pass around `allServices`, but I don't see how that would be an improvement.

I did think about adding a method similar to `lookupType` onto each entity, maybe `lookupMethod`. But to implement that, we still need `allComponents` passed to each entity.

So I just don't understand why you want to remove this necessary functionality - without it, the only way we can implement effects is to make users serialise the messages themselves.

andreaTP: The `serializeEffect` function already takes the `method` in the signature:
```js
serializeEffect(method, message, metadata)
```
Asking the user to pass the serializer doesn't look like a big price to pay IMO!

Passing around `allComponents` makes the code harder to be understood and tested/executed in isolation (e.g. breaks single responsibility), at least to my very fresh eyes on this codebase.

Just to be clear, I don't want to remove any necessary functionality, I'm only trying to find ways to evolve the codebase and I greatly appreciate your suggestions and explanations @jroper .

jroper: Yes, but the `method` that we expect users to pass does not contain a reference to the serialiser, which is why `allComponents` is needed. I actually think an improvement would be to not require users to even pass a descriptor, just the fully qualified name of the method would be good:

```javascript
.effect("example.MyService.MyMethod", {"foo":"bar"})
```

This would make it even easier I think for JavaScript users to use effects.

And asking a user to pass a serializer I think is very inappropriate. These APIs are used by a users business logic. It's the job of a framework (like this SDK) to ensure, as much as possible, that concerns such as serialization do not appear mixed with a users business logic. So, requiring that would be a regression in terms of what this SDK is supposed to achieve.

andreaTP: @jroper this proposed approach doesn't allow to get rid of `allComponents`.

Another design might be that each component exposes a method:
```
getEffectHandler("methodName")
```

so that the user code would look like:
```
.effect(otherComponent.getEffectHandler("example.MyService.MyMethod"), {"foo":"bar"})
```

johanandren:
Let's close this for now and possibly revisit in the future

PR #42, opened 2021-06-17
Refactor command helper

andreaTP:
I struggled to do this refactoring, unfortunately, but basically, here I confine calls to `this.call` into `handleCommand` solely and the rest are functions returning results.

This "should" help in getting the async handlers to work.

### DIFF
--- a/sdk/src/command-helper.js
+++ b/sdk/src/command-helper.js
@@ -53,6 +53,9 @@ class CommandHelper {
       if (err.failure && err.failure.commandId === command.id) {
         this.call.write(err);
         this.call.end();
+      } else {
+        console.error(err);
+        throw err;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


sdk/src/command-helper.js
```diff
@@ -77,37 +98,29 @@ class CommandHelper {
 
           ctx.streamed = command.streamed;
 
-          this.invokeHandler(() => handler(deserCommand, ctx), ctx, grpcMethod, reply => { return {reply}; }, "Command");
+          const reply = this.invokeHandlerLogic(() => handler(deserCommand, ctx), ctx, grpcMethod, "Command");
 
+          if (reply && reply.reply) {
```
andreaTP, line 103: I haven't managed to get rid of those lines, any suggestion is welcome, anything I have tried was breaking the TCK.

andreaTP:
Included in #43

PR #43, opened 2021-06-18
Support async user functions

andreaTP:
Based on the refactoring in #42 finally got it passing the tests! 🎉

### DIFF
--- a/sdk/integration-test/integration-testkit-test.js
+++ b/sdk/integration-test/integration-testkit-test.js
@@ -29,14 +29,51 @@ action.commandHandlers = {
   }
 };
 
+const value_entity = new akkaserverless.ValueEntity("./test/example.proto", "com.example.ExampleServiceTwo", "value-entity-example-service");
+
+value_entity.setInitial(entityId => (value_entity.lookupType("com.example.Example").create({})));
+
+value_entity.commandHandlers = {
+  DoSomethingOne: input => {
+    return {field: "ValueEntity Received " + input.field};
+  },
+  DoSomethingTwo: input => {
+    return new Promise((resolve) => {
+      setTimeout(() => resolve({field: "ValueEntityAsync Received " + input.field}), 1000);
+    });
+  }
+};
+
+const entity = new akkaserverless.EventSourcedEntity("./test/example.proto", "com.example.ExampleServiceThree", "event-sourced-entity-example-service");
+
+entity.setInitial(entityId => (entity.lookupType("com.example.Example").create({})));
+
+entity.setBehavior(e => {
+  return {
+    commandHandlers: {
+      DoSomethingOne: input => {
+        return {field: "EventSourcedEntity Received " + input.field};
+      },
+      DoSomethingTwo: input => {
+        return new Promise((resolve) => {
+          setTimeout(() => resolve({field: "EventSourcedEntityAsync Received " + input.field}), 1000);
+        });
+      }
+    },
+    eventHandlers: {}
+  };
+});
 
 const testkit = new akkaserverless.IntegrationTestkit({
   descriptorSetPath: "integration-test/user-function.desc",
 });
 
 testkit.addComponent(action);
+testkit.addComponent(value_entity);
+testkit.addComponent(entity);
 
 describe("The AkkaServerless IntegrationTestkit", function() {
+
   this.timeout(60000);
   before(done => testkit.start(done));
   after(done => testkit.shutdown(done));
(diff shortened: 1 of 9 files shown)


tck/event-sourced-entity.js
```diff
@@ -65,7 +65,10 @@ function process(request, state, context) {
     effects.forEach(effect =>
       reply.addEffect(two.service.methods.Call, { id: effect.id }, effect.synchronous)
     )
-  return reply
+
+  return new Promise((resolve, reject) => {
```
andreaTP, line 69: this test that the mechanism is working as expected (note that you can increase the timeout and the tests are still passing)

tck/value-entity.js
```diff
@@ -55,7 +55,10 @@ function process(request, state, context) {
   effects.forEach(effect =>
     reply.addEffect(two.service.methods.Call, { id: effect.id }, effect.synchronous)
   )
-  return reply
+
+  return new Promise((resolve, reject) => {
+    setTimeout(() => resolve(reply), 10);
+  });
```
johanandren, line 61: I think it would be good if we could cover both async and non async in the TCK across all component callback types somehow. Or wdyt @andreaTP is the paths the same so it does not really matter? 

I'm thinking with only async we could easily end up break the logic allowing us to just return a reply directly in all these places without noticing (even if that works in this PR.


  I guess it can't be a single run because the TCK itself does not support that but, we have complete control over the runner `run.js` so that could either run both or take a flag or something?

johanandren (commented):
I think a bunch of APIdocs needs update telling the users they can return a promise as well?

tck/value-entity.js
andreaTP: I changed just a couple of handlers to have integration tests for this feature; do you think I should revert?

There are still `commandHandlers` using the sync interface (that boils down to the same code-path now) e.g.:
https://github.com/lightbend/akkaserverless-javascript-sdk/blob/ab9f14f7adba43843710c7190e822ac6611d6a71/tck/value-entity.js#L72

sdk/src/command-helper.js
```diff
@@ -43,22 +43,43 @@ class CommandHelper {
    * @param command The command to handle.
    * @private
    */
-  handleCommand(command) {
+  async handleCommand(command) {
+    try {
+      const reply = await this.handleCommandLogic(command);
+      if (reply !== undefined) {
+        this.call.write(reply);
+      }
```
johanandren, line 51: But won't it will actually be `{ reply: undefined }` now because of how you changed the undefined check on the reply?

johanandren (commented):
Isn't something missing for the streaming out case, look at for example `action.js` in the tck and the `processStreamedOut` - we want to support imperative `ctx.reply` since it shouldn't be limited to 1:1 on incoming messages with streamed messages out, but it should probably allow 1:1 reply (or even imperative ctx.reply) with a promise as well?

Perhaps we can sort that out separately from this PR.

tck/value-entity.js
johanandren: Ok, hmm, I'd sleep better with a complete coverage (it would also somewhat document where it is supported and where it is still not possible to return a promise). On the other hand that might be a lot of extra work, so maybe this is enough, not quite sure.

sdk/src/command-helper.js
andreaTP: This reflects the check previously done here:
https://github.com/lightbend/akkaserverless-javascript-sdk/pull/43/files#diff-87e082fc96a33ca702e912a13787c4a08c23012e221b51cfe101985095b16e8fL190
that comes after the additional wrapping performed in `createReply`, if I understand it correctly it's good like this.

johanandren: You rewrap it in this change though: https://github.com/lightbend/akkaserverless-javascript-sdk/pull/43/files#diff-87e082fc96a33ca702e912a13787c4a08c23012e221b51cfe101985095b16e8fR106 so it won't be the return value that is undefined but a field of it?

andreaTP: @johanandren I might actually be wrong here but the wrapping was already done here:
https://github.com/lightbend/akkaserverless-javascript-sdk/pull/43/files#diff-87e082fc96a33ca702e912a13787c4a08c23012e221b51cfe101985095b16e8fL80

```js
reply => { return {reply}; }
```

In general, looking at the code, if we reach that point `reply` is always defined and the check might be redundant.
Should I try to remove the check to validate?

tck/value-entity.js
andreaTP: I added a few integration tests to cover this change:
https://github.com/lightbend/akkaserverless-javascript-sdk/pull/43/commits/2471b51538339dcf72f06026314768d34e256a0b

Thanks for the push @johanandren , do you think that this is enough or we should cover more?

sdk/src/command-helper.js
johanandren: Hmm, yea try that and see if things fail.

andreaTP: Removed the check, and I'm convinced now that is the right thing to do.
In case of undefined something will throw and it will be caught by the surrounding `try` block.

andreaTP:
> Isn't something missing for the streaming out case

possible, the intention was to cover the `commandHandler` user functions for Entities.

andreaTP:
@johanandren ready for another round of review 🙂

sdk/integration-test/integration-testkit-test.js
```diff
@@ -29,14 +29,51 @@ action.commandHandlers = {
...
+entity.setBehavior(e => {
+  return {
+    commandHandlers: {
+      DoSomethingOne: input => {
+        return {field: "EventSourcedEntity Received " + input.field};
+      },
+      DoSomethingTwo: input => {
+        return new Promise((resolve) => {
+          setTimeout(() => resolve({field: "EventSourcedEntityAsync Received " + input.field}), 1000);
+        });
```
johanandren, line 60: One thought, not sure we need to deal with it in this PR: One thing that could be nice to align with the rest of the api would be to have a convenience factory `Reply.promise`, since all kinds of other ways of replying and side effects etc is done that way. By having that a user can discover what is possible to return by auto completing on `Reply.` (in addition to looking at the JSdoc).

Perhaps a bit weird that you'd want to be able to also resolve the promise itself with `Reply` though `Reply.message().withSideEffects().whatNot()`

andreaTP: If you agree we can follow up on this good suggestion in a separate PR.
I think I would prefer to have first a consistent API experience for entities / value-entities etc.

johanandren: Yeah, lets defer that and do it separately.

johanandren (approved):

Issue #32, opened 2021-06-15
Deprecated warning when installing SDK


octonato:
We get a few deprecated warnings when running

```shell
npm install @lightbend/akkaserverless-javascript-sdk@0.7.0-beta.9
```

```node
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated grpc@1.24.10: This library will not receive further updates other than security fixes. We recommend using @grpc/grpc-js instead.
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
```

johanandren:
I wonder if these warnings has been fixed by the switch to grpc-tools in #38 @andreaTP ?

andreaTP:
Unfortunately, the switch to `grpc-js` resolves just one of the warnings.

> request@2.88.2

is still there but doesn't spill to user-land. (e.g. is used to download protoc)

> har-validator
> uuid

are transitive dependencies of the first.

andreaTP:
fixed in `0.7.0-beta.12`

PR #44, opened 2021-06-21
Run npm audit in CI

andreaTP:
Related to #32 
Do we aim for a stricter check?
e.g. should I remove `--omit dev` ?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
       - run:
           name: "check vulnerabilities in transitive dependencies"
           command: |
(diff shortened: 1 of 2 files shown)


johanandren (commented):
Not familiar with it since before but that sounds like it makes sense, I guess this PR would have to fix any current warnings as well though?

andreaTP:
> I guess this PR would have to fix any current warnings as well though?

There are none in the "production" dependencies, fortunately 😄

johanandren (approved):

pvlugter:
👍🏼  good idea to have this checked regularly

PR #46, opened 2021-06-22
Implementation of user function health check method

johanandren:
Impl of new health check method in discovery service, depends on lightbend/akkaserverless-framework#759


pvlugter:
Need to include this in the protocol update in #89, as the discovery protocol is typed now.

PR #47, opened 2021-06-22
Ts baby steps

andreaTP:
This is an initial setup to be able to work with TS in this codebase.

I have had to remove a few JSDoc annotations to get it working (thanks @janory !).

### DIFF
--- a/sdk/src/replicated-data/ormap.js
+++ b/sdk/src/replicated-data/ormap.js
@@ -74,22 +74,6 @@ function ORMap() {
    * using default values, the get method should not be used in queries where an empty value for the Replicated Data
    * means the value is not present.
    *
-   * @callback module:akkaserverless.replicatedentity.ORMap~defaultValueCallback
-   * @param {module:akkaserverless.Serializable} key The key the default value is being generated for.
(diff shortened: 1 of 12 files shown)


sdk/src/replicated-data/ormap.js
```diff
@@ -63,22 +63,6 @@ function ORMap() {
...
-   * If this returns a Replicated Data object, it will be added to the map.
-   *
-   * Care should be taken when using this, since it means that the get method can trigger elements to be created. If
-   * using default values, the get method should not be used in queries where an empty value for the Replicated Data
-   * means the value is not present.
-   *
-   * @callback module:akkaserverless.replicatedentity.ORMap~defaultValueCallback
-   * @param {module:akkaserverless.Serializable} key The key the default value is being generated for.
-   * @returns {undefined|module:akkaserverless.replicatedentity.ReplicatedData} The default value, or undefined if no default value should be returned.
-   */
```
johanandren, line 80: Explanations for how things work are lost now, I think we still need them but in some other form.

pvlugter: 👍🏼

PR #48, opened 2021-06-22
Ts baby steps

andreaTP:
Adding support for writing TypeScript in this code-base.


johanandren (commented):
I think we still want a lot of those removed docs, but elsewhere if the current typedef is not allowed with ts.

andreaTP:
@johanandren managed to just move around JSDocs so that they are not breaking TS compilation! 🎉 

Adjusted the build and the CI accordingly.

pvlugter (approved):
Seems good to me 👍🏼

PR #53, opened 2021-06-24
docs: tags for reference in proto.adoc

ennru:
Add Antora tags to source snippets for developer.lightbend.com/docs/akka-serverless/javascript/proto.html from validated Javascript SDK sources.﻿

### DIFF
--- a/samples/js-valueentity-shopping-cart/proto/shoppingcart_api.proto
+++ b/samples/js-valueentity-shopping-cart/proto/shoppingcart_api.proto
@@ -75,8 +75,8 @@ service ShoppingCartService {
   }
 
   rpc RemoveItem (RemoveLineItem) returns (google.protobuf.Empty) {
-    option (google.api.http).post = { // <2>
-            "/cart/{cart_id}/items/{product_id}/remove"
+    option (google.api.http) = { // <2>
+            post: "/cart/{cart_id}/items/{product_id}/remove"
         };
   }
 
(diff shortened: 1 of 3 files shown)


samples/js-valueentity-shopping-cart/proto/shoppingcart_api.proto
```diff
@@ -14,14 +14,21 @@
...
 
+// tag::proto_package[]
 package com.example.shoppingcart;
+// end::proto_package[]
 
-import "google/protobuf/empty.proto";
+// tag::proto_imports[]
 import "akkaserverless/annotations.proto";
 import "google/api/annotations.proto";
+import "google/protobuf/empty.proto";
```
octonato, line 28: When changing the Sunglasses UI, I ran into an issue where the ordering of imports matter. The akkaserverless import needed to come after the google annotations one. 

I think this was more a problem of grpc-web, because this ordering is known to work here. But perhaps good to change to avoid issues. 

```suggestion
import "google/api/annotations.proto";
import "google/protobuf/empty.proto";
import "akkaserverless/annotations.proto";
```

octonato (approved):
LGTM

samples/js-valueentity-shopping-cart/proto/shoppingcart_api.proto
ennru: Wow.

PR #52, opened 2021-06-24
Merge https://github.com/lightbend/akkaserverless-npm-js

andreaTP:
The entire CI has been ported to Circle and is running.
Haven't tested the publishing yet, but we can always fix it on top in case is required.

Issue #30, opened 2021-06-11
[epic] Bring in npm-js and javascript codegen


octonato:
As discussed, we plan to merge into this project the npm-js and the javascript codegen. 

The execution plan is as follow: 

## Phase 1:

- [ ] **move**: bring `npm-js` under `javascript-sdk` (preserving commit history)
- [ ] **build**: must be integrated in CircleCI (similar to what we are doing to the samples already)
- [ ] **versioning**: we want the `npm-js` to have the same version of the `javascript-sdk`. 
- [ ] **publishing**: `npm-js` artifact must be published under a new name because of version misalignment 
- [ ] **docs**: documentation must be adapted since we will have a new artifact name
- [ ] **github issues**: move all open issues from `npm-js` to `javascript-sdk`

## Phase 2:

First we need to split codegen into codegen-core, codegen-java and codegen-js (preserving commit history) See https://github.com/lightbend/akkaserverless-codegen/issues/147

- [ ] move `codegen-js` under `javascript-sdk` (preserving commit history)

Please, feel free to add here any other point of action that you believe we will need.

octonato:
To merge a project preserving the history, we should do the following:

How to merge a git repo (foo) into a repo (bar):

* **foo**: move the contents of the repo into a directory (make sure that includes any dot files, such as .gitignore, but then make sure you don't nest the .git directory)
* **foo**: remove whatever you don't want to merge in bar (eg. CI build files)
* **foo**: commit locally
* **foo**: optional - remove issues and PR references from commit messages
  * needs git-filter-repo
  * run `git-filter-repo --message-callback 'return re.sub(b" \(#\d+\)", b"", message)'`
  * run `git-filter-repo --message-callback 'return re.sub(b"#\d+", b"#XXX", message)'`
* **bar**: add foo as a remote - use path in your file system, you don't need to push the changes to foo origin
* **bar**: fetch the branch in foo remote you want to merge
* **bar**: git merge --allow-unrelated-histories foo/branch

Credits to @dwijnand for the git fu and to @pvlugter for the `git-filter-repo` trick.

pvlugter:
I think this is all done and can be closed now.

PR #55, opened 2021-06-24
Merge codegen and npm-js

andreaTP:


andreaTP:
This is already on top of #52 .
The publishing and re-use of the local codegen is still missing, but all the rest is already fine (sbt test are already in Circle as well)

pvlugter (changes requested):
Difficult to review something like this, but some suggestions:
 
- We want to keep all the history, so we'll be merging with a merge commit (or possibly a rebase and merge), not a squash and merge. So it would be useful to tidy up the extra commits that have been added (like `test in circleCI`, `more`, `fix`, and so on) so we have as clean a history as possible. It's a shame codegen projects were using merge commits already, rather than squash and merge.
- Rather than having all the earlier commits and then moving their location in an extra commit, rewrite the history now with path renames. Can use `git-filter-repo` for this. For example: `git-filter-repo --path-rename some-path/:codegen/some-path/`. Or even just move everything first when applying to the separate repo, with something like: `git-filter-repo --path-rename :codegen/`. That will make the merged history much easier to follow, as if it was always under `codegen` separately.
- Looks like some commits are not connected to a github user (failing CLA validator) so let's rewrite those commits to have a connected author email.
- I think it's fine to first do the merge with all the code, with renamed paths in the history (as suggested above), and disabled CI, then get everything working in separate pull requests (like #56).

andreaTP:
@pvlugter :

- opened #58 with strictly only the merge of folders
- I followed those instructions: https://github.com/lightbend/akkaserverless-javascript-sdk/issues/30#issuecomment-866133909 if you prefer something to be done differently please go ahead directly with a PR and I will decline the rest
- Not much I can do about those commits other than asking @RoryStokes if he can make sure that all the emails he is committing from are connected to his GH account.
- I'll follow up stacking the rest of the needed changes as soon as we start to merge something

pvlugter:
@andreaTP, I've created a draft of just the npm-js commits in #61, looking to create a cleaner history:
- renamed paths in history to be under `npm-js` and added `npm-js` prefix to all commits
- rebased to remove all the merge commits and have just one linear history
- rewrote Rory's disconnected email address

pvlugter:
Similarly created a cleaner history for codegen in #62.

andreaTP:
Included in #63

PR #56, opened 2021-06-28
Publish native binaries

andreaTP:


andreaTP:
I attempted but I can't really go any further without major trouble and without merging this and #55 .

My plan would be to:
 - merge everything
 - do a release
 - fix on top whatever is left
 - integrate more tightly e2e tests using the locally built `codegen`
 
 It would be nice if someone can assist me during the process.

RayRoestenburg (approved):

pvlugter (commented):
LGTM.

Looks like a fairly complicated setup, with native images per platform, but when this is for javascript.

Have we changed the artifact names (or similar) now that we've aligned the versions with the SDK version? We'll need to make sure we don't end up with version conflicts at some point.

pvlugter:
Sounds good to first tidy up and merge #55, then get things working and fully integrated on top of that, incrementally.

andreaTP:
Included in #63

PR #57, opened 2021-06-28
Use types from JSDocs for now

andreaTP:
This replaces the TS generated types with the one from JSDocs

### DIFF
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -79,7 +79,7 @@
   },
   "scripts": {
     "clean": "rm -rf ./dist",
-    "copy-files": "cp -r ./proto/ ./dist/proto",
+    "copy-files": "cp -r ./proto/ ./dist/proto && cp index.d.ts ./dist",
     "compile": "tsc && npm run copy-files",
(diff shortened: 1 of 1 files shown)


sdk/package.json
```diff
@@ -79,7 +79,7 @@
   },
   "scripts": {
     "clean": "rm -rf ./dist",
-    "copy-files": "cp -r ./proto/ ./dist/proto",
+    "copy-files": "cp -r ./proto/ ./dist/proto && cp index.d.ts ./dist",
```
janory, line 82: The `index.d.ts` file is generated during the `prepare` step, but the `compile` doesn't depend on that step.
Wouldn't make sense to do the copy in the `prepare.sh` script?

janory (approved):

sdk/package.json
andreaTP: the compile is executed after the `prepare.sh` ATM and the result is going to be overridden by `compile`.
But I'm open to change (maybe in another PR) the order/sequence of commands if you have a good proposal @janory 🙂

PR #58, opened 2021-06-29
Merge again

andreaTP:
This only moves the two separate repos to this codebase in folders.


andreaTP:
Done by @pvlugter in #61 and #62

PR #59, opened 2021-06-29
Remove Cloudevent conversions

andreaTP:
Looks like that this file is here for historical reasons.

Please let me know if I'm missing something!


janory (approved):
LGTM! 👍

pvlugter:
I'd think that the CloudEvents conversion was there as a convenience, for whenever events do have CloudEvents metadata, so not sure why we want to remove a utility like this. I'll let @jroper comment on the original intentions.

jroper:
This shouldn't have been removed. Akka Serverless supports CloudEvents. We propagate and/or generate CloudEvent metadata for all eventing calls. This metadata is passed through our own Metadata abstraction, with all CloudEvent fields being prepended with `ce-`, which is a common CloudEvent convention. Then, at least the intention was in the SDKs that we would offer a convenient high level API for accessing this metadata, rather than requiring users to know exactly how CloudEvent was exposed in our metadata - an implementation detail that they shouldn't be exposed to. The high level API should also return the correct CloudEvent type - so for example, the one call that has been left is `getSubject`, in CloudEvents, it's a string, our API should return a string.

PR #60, opened 2021-06-29
More TypeScript

andreaTP:


sdk/src/reply.ts
```diff
@@ -0,0 +1,188 @@
...
+    readonly metadata: Metadata,
+  ) {}
+}
+
+/**
+ * A return type to allow returning forwards or failures, and attaching effects to messages.
+ *
+ * @memberOf module:akkaserverless.replies
+ */
+export class Reply {
```
pvlugter, line 43: Looks like the typed version of Reply will be useful 👍🏼

sdk/test/metadata.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    meta.set('key1', 'hello1');
+    meta.set('key1', 'hello2');
+
+    // Act
+    const res = meta.get('key1');
+
+    // Assert
+    expect(res.length).to.be.equal(2);
+    expect(res[0]).to.be.equal('hello1');
+    expect(res[1]).to.be.equal('hello2');
```
pvlugter, line 57: Expect that chai matchers can also match the whole array to check same length and members? This also looks fine.

pvlugter (approved):
LGTM. Nice to see more typing, and more tests 👍🏼

PR #61, opened 2021-06-30
Merge npm-js

pvlugter:
Trying a different approach for merging in the `npm-js` codebase, to get a cleaner history. This is only `npm-js` and not codegen js stuff currently.

Paths in history have been rewritten to all be under `npm-js` subdirectory, and commit messages prefixed with `npm-js`.

Rebased to remove the merge commits from the original history, with committer and commit date details rewritten to match original author details and dates. Rebased on current `main` branch so that merging with unrelated histories is not needed.

And updated Rory's disconnected email addresses to be his current email address.

~Commits and files have otherwise been left as is. For example, there's still `npm-js/.github/workflows/` directory with old CI setup.~ Have filtered out the CI build files. After merge we can add the integration commits from #55.

Expect CI to fail for things like licence checks, which we can update with additional commits.

If we merge with a merge commit, we should be able to retain all the original commits but with a simpler merged history.


pvlugter:
**NOTE: don't squash and merge this**

andreaTP (approved):

PR #62, opened 2021-06-30
Merge codegen

pvlugter:
Similar process to #61, for getting a cleaner history for merging codegen.

The java codegen has been filtered out, while core codegen is kept. The build still refers to `codegen-java` which will need to be updated. I think the plan is to include codegen core in the SDKs currently, rather than as a dependency. We can otherwise filter out the core as well, and then update the build.

Rebased to remove merge commits. Commits all prefixed with `codegen`. Rebased on current `main` branch to create related history.


pvlugter:
**NOTE: don't squash and merge this**

andreaTP (approved):

PR #63, opened 2021-06-30
Integrate prjs

andreaTP:
```yaml
professional:
    yaml:
        engineer: [👓 ]
```


codegen/set-version.sh
```diff
@@ -0,0 +1 @@
+echo "ThisBuild / version := \"$(../sdk/bin/version.sh)\"" > version.sbt
```
pvlugter, line 1: Could also use an sbt-dynver setup that matches the version script, but also good to use the one script everywhere.

pvlugter (approved):
LGTM. Fairly complicated setup for the native image building and publishing, thanks for yaml'ing your way through that.

codegen/set-version.sh
andreaTP: Let me finalize [this](https://github.com/dwijnand/sbt-dynver/pull/188) PR and we are going to be consistent 🙂

PR #64, opened 2021-06-30
Remove un-necessary dependencies

andreaTP:
Follow up from:
https://typesafe.slack.com/archives/CQUASEUDV/p1625068565235800

### DIFF
--- a/npm-js/create-akkasls-entity/template/base/package.json
+++ b/npm-js/create-akkasls-entity/template/base/package.json
@@ -7,11 +7,7 @@
     "npm": ">=6.0.0"
   },
   "dependencies": {
-    "@grpc/grpc-js": "^1.3.0",
-    "@grpc/proto-loader": "^0.1.0",
-    "@lightbend/akkaserverless-javascript-sdk": "0.7.0-beta.11",
-    "google-protobuf": "^3.0.0",
-    "protobufjs": "^6.11.2"
+    "@lightbend/akkaserverless-javascript-sdk": "0.7.0-beta.13"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


samples/valueentity-counter/package.json
```diff
@@ -7,11 +7,7 @@
     "npm": ">=6.0.0"
   },
   "dependencies": {
-    "@grpc/grpc-js": "^1.3.0",
```
andreaTP, line 10: Note for the reviewers, this project is built and tested in CI, so we are confident that the dependencies are not needed.

npm-js/create-akkasls-entity/template/base/package.json
```diff
@@ -7,11 +7,7 @@
     "npm": ">=6.0.0"
   },
   "dependencies": {
-    "@grpc/grpc-js": "^1.3.0",
-    "@grpc/proto-loader": "^0.1.0",
-    "@lightbend/akkaserverless-javascript-sdk": "0.7.0-beta.11",
-    "google-protobuf": "^3.0.0",
-    "protobufjs": "^6.11.2"
+    "@lightbend/akkaserverless-javascript-sdk": "0.7.0-beta.13"
```
pvlugter, line 10: Side note: we probably want this version to be updated automatically at some point...

pvlugter (approved):
LGTM

I think some of the samples had explicit dependencies from when they depended on the SDK directly as a path, which is probably where this originated.

npm-js/create-akkasls-entity/template/base/package.json
andreaTP: yes, that work is not finalized yet. Going to work on it.

PR #65, opened 2021-06-30
introduce eslint for TS development

andreaTP:
Apply it to `.js` files is not really useful, since, at the moment, there are way too many violations.
Instead is a good check for new TS files (already fixed a few issue that arose.


pvlugter (approved):
LGTM

Good idea to have eslint enabled now

PR #66, opened 2021-07-01
Add doc sources for JavaScript SDK from main docs

pvlugter:
Similar to https://github.com/lightbend/akkaserverless-java-sdk/pull/70.

Copy the JavaScript SDK docs from the main Akka Serverless docs, so they can be maintained here.

The partials from the developing module are still partials here (but under javascript module so they don't conflict), in case we want to move these to a common repo for all SDK docs later on.

Cross-references to the main docs are all converted to external links to the published production docs, so the docs here can be used and checked separately. These are full links in the asciidoc sources, so they get validated by the asciidoc external link checker, which doesn't support variables.

There's a simple setup for building and viewing the docs locally — uses a default theme, as the Akka Serverless doc theme is defined as supplemental ui in a private repo currently.

This will be published to `docs/tmp/current` branch during the transition, until sources actually removed from the main docs, to avoid duplicate sources.

Other than fixing links, haven't altered these docs. The examples needed proto files from the java eventing shopping cart sample, which I've copied in. The examples should all be tested, which we can follow up on, along with any other improvements.


pvlugter:
Published to `docs/tmp/current` and sources removed from main docs in https://github.com/lightbend/akkaserverless-docs/pull/748.

andreaTP (approved):

PR #68, opened 2021-07-01
Tighten up your versions

andreaTP:
This PR makes it possible to run full e2e tests without external dependencies so that all the versions can stay aligned now.

It should be closer to a working publishing step as well.


pvlugter (approved):
Nice, LGTM

johanandren (approved):

PR #77, opened 2021-07-01
Akkaserverless service and integration testkit in TS

andreaTP:
Akkaserverless and Integration Testkit are now in TS.
There are still a few `any` around to keep compatibility with the old JS API, but I think it's an improvement.

Note:
I gitignored the `package-lock` of the `tck` because, using Typescript, we should anyhow always compile the `sdk` (e.g. we cannot import directly from the directory)


andreaTP:
Need to port `integration-testkit` to TS to make it work probably, will do it tomorrow 🙂

sdk/src/akkaserverless.ts
```diff
@@ -0,0 +1,454 @@
...
+import * as settings from '../settings';
+
+import * as discovery from '../proto/akkaserverless/protocol/discovery_pb';
+import * as discovery_grpc from '../proto/akkaserverless/protocol/discovery_grpc_pb';
+import * as google_protobuf_empty_pb from 'google-protobuf/google/protobuf/empty_pb';
+import { PackageInfo } from './package-info';
+
+export class Bindings {
+  readonly address: string;
+  readonly port: number;
```
pvlugter, line 28: Not important, but not sure `Bindings` reads that well. Maybe call the overall class `Address` with `host` and `port` fields/parameters? (also aligned with the env vars)

pvlugter (approved):
LGTM

sdk/src/akkaserverless.ts
```diff
@@ -0,0 +1,454 @@
...
+
+export interface Component {
+  serviceName?: string;
+  desc?: string | string[];
+  service?: any;
+  options?: ComponentOptions;
+  grpc?: grpc.GrpcObject;
+  componentType?: () => string;
+  register?: (components: any) => Service;
+}
```
johanandren, line 96: I think that for these two types, there's a few fields that should not be optional: `serviceName`, `entityType`, `options` for sure, maybe more.

johanandren (commented):
Looks good, added some smaller feedback

sdk/src/akkaserverless.ts
andreaTP: agree 👍

sdk/src/akkaserverless.ts
```diff
@@ -0,0 +1,454 @@
...
+export interface Service {
+  componentType: () => string;
+  register: (server: grpc.Server) => void;
+}
+
+export interface EntityPassivationStrategy {
+  timeout?: number;
+}
+
+export interface ComponentOptions {
```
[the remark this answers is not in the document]
andreaTP: In this case, we can't directly rely on the Protocol types, since additional fields are included ( e.g. `includeDirs`). Not judging the current design, just making this PR as less breaking as possible...

sdk/src/akkaserverless.ts
```diff
@@ -0,0 +1,454 @@
...
+  serviceName?: string;
+  desc?: string | string[];
+  service?: any;
+  options?: ComponentOptions;
+  grpc?: grpc.GrpcObject;
+  componentType?: () => string;
+  register?: (components: any) => Service;
+}
+
+class DocLink {
```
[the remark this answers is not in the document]
andreaTP: yup, is not exposed

sdk/src/akkaserverless.ts
```diff
@@ -0,0 +1,454 @@
...
+
+    Object.values(componentTypes).forEach((services: any) => {
+      services.register(this.server);
+    });
+
+    const discoveryServer = this.getDiscoveryServer();
+
+    this.server.addService(discovery_grpc.DiscoveryService, discoveryServer);
+
+    const that = this;
```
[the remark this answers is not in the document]
andreaTP: moved to its own method and looks much nicer

sdk/src/akkaserverless.ts
andreaTP: this is actually for keeping compatibility with the existing tests and the precedent JS encoding.

sdk/src/akkaserverless.ts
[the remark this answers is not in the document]
andreaTP: correct

andreaTP:
@johanandren I think it's ready for next round 🙂

sdk/src/akkaserverless.ts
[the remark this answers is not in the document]
johanandren: Ok, I think I misinterpreted this to be about the duck-typey objects in the old discovery logic where it was internal but it's rather is a way to pass the options from components to the discovery?

sdk/src/akkaserverless.ts
johanandren: I don't get that, those are not optional values in the JS component impls, they are always present.

andreaTP: well, in JS everything is optional ... anyhow removed a few `?`

johanandren (approved):

sdk/src/akkaserverless.ts
```diff
@@ -0,0 +1,439 @@
...
+      },
+      proxyTerminated(
+        call: grpc.ServerUnaryCall<
+          google_protobuf_empty_pb.Empty,
+          google_protobuf_empty_pb.Empty
+        >,
+        callback: grpc.sendUnaryData<google_protobuf_empty_pb.Empty>,
+      ) {
+        callback(null, new google_protobuf_empty_pb.Empty());
+      },
```
pvlugter, line 335: This no longer runs the actual `proxyTerminated` logic now?

andreaTP: good catch @pvlugter ! I'll double-check and fix it on top!

PR #89, opened 2021-07-06
bump: update to proxy and protocol 0.7.0-beta.11

pvlugter:
Resolves #88. Includes the bump to 0.7.0-beta.11 from #86. Update the replicated data types in line with the new protocol for replicated entities.

PR #90, opened 2021-07-06
build: compile SDK protocol directly for codegen

pvlugter:
Codegen shouldn't be depending on the Java SDK for the akkaserverless annotations. Compile the SDK protocol directly instead, with the akkaserverless version for the protocol/proxy.

Also including in #89 to check the bump to the new protocol.


andreaTP:
Isn't it included in #89 ?

pvlugter:
Yep, also included in the protocol bump. This was uncovered by updating the dependencies, not directly related to it, so added separately in case the update had more changes needed.

I might just merge this first separately, so the two changes are separate.

PR #91, opened 2021-07-06
Remove start stop on entities

andreaTP:
This reduces the API and improves the separation of concerns.


johanandren (approved):

pvlugter:
Could be useful to check if any samples are using this entity start shortcut and update them. But it will be obvious anyway with it removed.

PR #93, opened 2021-07-06
[RFC] Bubble serialization up

andreaTP:
I applied this change just to `action` since is the most tested and shows the intention.

Here we:
 - remove the proto logic in the `Component`
 - inject the serializer (still through the `Component` to keep the constructor signature ... for now)
 - use the injected `serializationSupport` for everything
 
It's still not a "final design", but a step forward, and we need to do more cleanup and write more tests before applying the same pattern to the other components.


retgits:
For a user of the SDK, will this change anything in the code they write?

andreaTP:
> For a user of the SDK, will this change anything in the code they write?

No

pvlugter (commented):
LGTM to move all the serialization support together.

Failing the integration testkit test.

sdk/src/serialization-support.js
```diff
@@ -0,0 +1,83 @@
...
+
+  getEffectSerializer() {
+    if (!this.allComponents) {
+      throw new Error('Please run `setComponents` first');
+    } else {
+      return new EffectSerializer(this.allComponents);
+    }
+  }
+
+  getGrpc() {
```
andreaTP, line 53: Having to expose `Grpc` and `Service` doesn't sounds great 🙁 
Probably this PR needs a few more steps to go through ...

johanandren:
Lets' close this for now and possibly revisit in the future

PR #115, opened 2021-07-08
add Leon's videos to docs

thothmoses:
per issue https://github.com/lightbend/akkaserverless-docs/issues/763


thothmoses:
This is for https://github.com/lightbend/akkaserverless-javascript-sdk/pull/115
I  still cannot build on the SDK Repo side.
Also, I want to build links into this from the main docs but this needs to be merged first. So that I can correctly assign the HTTP URLs

docs/src/modules/javascript/pages/value-entity-videos.adoc
```diff
@@ -0,0 +1,20 @@
+= Value entities and stateful services: Video 1-2-3
+
+The following videos will give you a quick overview of building a start service with value entities.
+
+First, a review of what Akka Serverless provides for your developers.
+
+link:https://www.youtube.com/watch?v=D9_JM2sRh98[Akka Serverless - the best of serverless and state without any hassle, role=yt-widget]
+
+Now, just view the three following videos. In less than five minutes you will cover the basics.
```
retgits, line 9: I'd recap a part of the `Simplified Programming Model` here, like

"In less than five minutes, you can build a stateful serverless service using Value Entities. These videos walk you through the three steps to you need to develop:
1. Define the API and the domain data model in gRPC descriptors.
2. Choose a state model (this is your Value Entity).
3. Implement your business logic in JavaScript."

docs/src/modules/javascript/pages/value-entity-videos.adoc
```diff
@@ -0,0 +1,20 @@
+= Value entities and stateful services: Video 1-2-3
+
+The following videos will give you a quick overview of building a start service with value entities.
```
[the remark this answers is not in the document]
thothmoses: Ack! ok - I'll make that change - brain fog

docs/src/modules/javascript/pages/value-entity-videos.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+First, a review of what Akka Serverless provides for your developers.
+
+link:https://www.youtube.com/watch?v=D9_JM2sRh98[Akka Serverless - the best of serverless and state without any hassle, role=yt-widget]
```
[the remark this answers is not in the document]
thothmoses: I agree - I shoehorned it in to get to see how other people felt about it.

docs/src/modules/javascript/pages/value-entity-videos.adoc
thothmoses: cool - adding now

docs/src/modules/javascript/pages/value-entity-videos.adoc
```diff
@@ -1,12 +1,10 @@
 = Value entities and stateful services: Video 1-2-3
 
-The following videos will give you a quick overview of building a start service with value entities.
+These videos will give you a quick overview of implementing Value Entities in Akka Serverless. In less than five minutes, you can build a stateful serverless service using Value Entities. These videos walk you through the three steps to you need to develop:
```
rstento, line 3: @retgits, don't you think it will be confusing that the steps in these videos are different than what is in the doc? In the doc, we suggest they design first and choose their state model. And, your second video isn't really about "choosing" the state model, I don't remember you even mentioning Event Sourced entities? It is about implementing in a .js file.

rstento (commented):
@thothmoses, the issue to add these videos was in the akkaserverless-docs repo, so I was surprised to see that you added them to this SDK repo. I'd really prefer them in the akkaserverless-docs repo and then link to them from the SDK repos. That's why I suggested adding the new page in the Tutorials section, which is in the docs repo. WDYT?

docs/src/modules/javascript/pages/value-entity-videos.adoc
thothmoses: Sorry, I thought you were just going through possibilities. When I looked at the tutorials there was no interstitial location where the videos seemed a natural fit. And the new page added after the tutorials felt a bit disarticulated, outside the flow. Whereas, the vid page seemed like an enhancement for the value entity material. But for ease of maintenance, I can see that this would be better off on the Doc Repo side. Let me know, I can move it in a trice.
Oh, BTW, Peter helped me out, and I now know how to build on the SDK side. Right now it uses the default Antora UI, but I think we can work to update that. At least we can validate now.

rstento: Great that you can build! What about in the Tutorial section before the IOT and shopping cart? It is a Hello World example in JavaScript, that's how I would name it.

thothmoses: Ok Great idea - I'll rebrand it and put it there as a 'Hello World'

PR #126, opened 2021-07-14
ci: fix java versions for macOS builds

pvlugter:
The macOS native image build failed for 0.31.0 release, as sbt was running with openjdk 16:

https://app.circleci.com/pipelines/github/lightbend/akkaserverless-javascript-sdk/426/workflows/64b38cfe-fd38-4729-bae4-488022d479de/jobs/2037

This must have changed at some point since last release, where sbt depends on brew-installed openjdk by default, and this is now JDK 16. There are a few ways to fix this. There is an AdoptOpenJDK 11 install with the OS already. And JAVA_HOME can be specified. I've opted to link openjdk@11 as the default openjdk, and not install the openjdk dependency for sbt, so that using the sbt command as is will run with the explicitly installed openjdk by default.

For 0.31.0, I've manually published the native image for macOS (from the CI machine, with rerun job with ssh).


octonato (approved):
LGTM

PR #127, opened 2021-07-14
feat: move write consistency to overall replicated entity settings

pvlugter:
Depends on https://github.com/lightbend/akkaserverless-framework/pull/795 being published first.

Move write consistency from a per-update setting to an overall replicated entity setting.


pvlugter:
Now included in #131 along with other changes.

PR #92, opened 2021-07-06
Test EventSerializer

andreaTP:
This is more preparation work for bigger refactorings.

PR #130, opened 2021-07-16
test: fix test setup so that they all run again

pvlugter:
Since #92 looks like we stopped running all the tests. https://github.com/lightbend/akkaserverless-javascript-sdk/pull/92/commits/46ba10723c1514b9bde4222dfb59a8e763fcd3ff ended up updating so that only the replicated data tests were [running](https://app.circleci.com/pipelines/github/lightbend/akkaserverless-javascript-sdk/380/workflows/27859c17-21a9-467c-9031-fd27f8c116e2/jobs/1697) -> `77 passing`.

This reverts some of the changes for compiling the test protobufs, and rather than trying to include particular tests, this uses mocha's exclude/ignore to not run compiled proto as tests.

I'll check this works in CI too, but all tests running again -> `149 passing`

### DIFF
--- a/sdk/.gitignore
+++ b/sdk/.gitignore
@@ -5,5 +5,3 @@ apidocs
 user-function.desc
 index.d.ts
 *.tgz
-test/example_pb.d.ts
-test/example_grpc_pb.d.ts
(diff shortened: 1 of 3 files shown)


pvlugter:
That runs all the tests again in CI as well 👍🏼

andreaTP (approved):
Thanks for fixing my mistakes! 😅

octonato (approved):

PR #132, opened 2021-07-16
Update views.adoc

### LINKED ISSUE #119: Rearrange Querying a View and Registering a View subsections.
@mckeeh3 commented on [Sun Jun 13 2021](https://github.com/lightbend/akkaserverless-docs/issues/694)

In Implementing Views in both [JavaScript](https://docs.akkaserverless.dev/js-services/views.html) and [Java](https://docs.akkaserverless.dev/java-services/views.html), move the Querying a View ([JS](https://docs.akkaserverless.dev/js-services/views.html#query), [Java](https://docs.akkaserverless.dev/java-services/views.html#query))  subsection after all of three main subsections (Value Entity, Event Source Entity, Topics).

At the end of each of the three main subsections, add links to the "Querying a View" subsection. Something like "See Querying a View for more about how to select data."

At the end of each of the three main "Creating a View from a X" subsections, add a registering subsection specific to that View type.

### LINKED ISSUE #120: [Feedback from user] Table names for Views
@retgits commented on [Wed Jul 07 2021](https://github.com/lightbend/akkaserverless-docs/issues/764)

Notes from NIHITO session

> How do we know what the table name is for a View?

After asking a bit more, it was quite unclear to the user how the table names for Views are used and created and how they refer to the entity it receives messages from.

---

@coreyauger commented on [Wed Jul 07 2021](https://github.com/lightbend/akkaserverless-docs/issues/764#issuecomment-875740041)

Can we not make the table name the same as the service name for the view.. and just remove it entirely from the user having to deal with it? 

---

@patriknw commented on [Wed Jul 07 2021](https://github.com/lightbend/akkaserverless-docs/issues/764#issuecomment-875777181)

> Can we not make the table name the same as the service name for the view.

We might want to support more than one table per view in the future.

Right now you can use any table name you like, as long as you use the same in the query and update method. There should be good error message if they don't match.

---

@rstento commented on [Wed Jul 07 2021](https://github.com/lightbend/akkaserverless-docs/issues/764#issuecomment-875781402)

I had the same question when I worked on the doc, that is why the doc explicitly says: "You can use any name for the table." but it is in the explanation of the Protobuf example, item 3, so could be easily overlooked. I think we should add this info to the introduction of Views, explaining that one step is to define the table name, and you can choose any name.

rstento:
Fixes #119, Fixes #120.

I will be doing a separate PR for #98, after I deal with review comments on this PR. 

@mckeeh3, to make it easier to review in context, I've uploaded a zip of the doc here: Just unzip it and view the page on Views in the JavaScript section. Otherwise it is difficult to envision how these changes impact the user experience.

### DIFF
--- a/docs/src/modules/javascript/pages/views.adoc
+++ b/docs/src/modules/javascript/pages/views.adoc
@@ -4,18 +4,28 @@ include::ROOT:partial$include.adoc[]
 include::partial$views.adoc[]
 
 [#value-entity]
-== View from Value Entity
+== View from a Value Entity
 
-Using an example of a customer registry, you can define a `Customer` in Protobuf as:
+Consider an example of a Customer Registry service with a `customer` Value Entity. When `customer` state changes, the entire state is emitted as a value change. Value changes update any associated Views. To create a View that shows customers by their name:
+
+. <<_define_the_view_service_descriptor>> for a service that selects customers by name and associates a table name with the View. The table is created and used by Akka Serverless to store the View, use any name for the table.
+. <<#register-view>>.
+
+This example assumes the following `customer` state is defined in a `customer_domain.proto` file:
 
 [source,proto,indent=0]
 ----
 include::example$js-customer-registry/customer_domain.proto[tag=domain]
 ----
 
-When `Customer` state changes, the entire state is emitted as a value change, which updates any associated Views.
+=== Define the View service descriptor
 
-To get a view of multiple customers by their name, define the View as a `service` in Protobuf:
+To get a view of multiple customers by their name, define the View as a `service` in Protobuf. The descriptor defines:
+
+* How to update the View
+* The source of View data
+* A `table` property that can be any name
+* The query that returns customers by name
 
 [source,proto,indent=0]
 ----
@@ -28,7 +38,9 @@ include::example$js-customer-registry/customer_view.proto[tag=service]
 <4> The `GetCustomers` method defines the query to retrieve a stream of `Customer`.
 <5> The `(akkaserverless.method).view.query` annotation defines that this method is used as a query of the View.
 
-If the query is supposed to return only one result, the `stream` can be removed from the return type:
+[NOTE]
+====
+If the query should return only one result, `stream` can be removed from the return type:
 
 [source,proto,indent=0]
 ----
@@ -38,64 +50,14 @@ include::example$js-customer-registry/customer_view.proto[tag=unary]
 <1> Without `stream` when expecting single result.
 
 When no result is found, the request fails with gRPC status code `NOT_FOUND`. A streamed call completes with an empty stream when no result is found.
+====
 
-[#query]
-== Query
-
-Define View queries in a language that is similar to SQL. The following examples illustrate the syntax.
-
-All customers without any filtering conditions (no WHERE clause):
-[source,proto,indent=0]
-----
-SELECT * FROM customers
-----
-
-Customers with a name matching the `customer_name` property of the request message:
-[source,proto,indent=0]
-----
-SELECT * FROM customers WHERE name = :customer_name
-----
-
-Customers with matching `customer_name` AND `city` properties of the request message:
-[source,proto,indent=0]
-----
-SELECT * FROM customers WHERE name = :customer_name AND address.city = :city
-----
-
-Customers in city matching a literal value:
-[source,proto,indent=0]
-----
-SELECT * FROM customers WHERE address.city = 'New York'
-----
-
-Possible filter predicates are:
-
-* `=` equals
-* `!=` not equals
-* `>` greater than
-* `>=` greater than or equals
-* `<` less than
-* `\<=` less than or equals
-
-The filter conditions can be combined with `AND`/`OR`.
-
-[source,proto,indent=0]
-----
-SELECT * FROM customers WHERE
-  name = :customer_name AND address.city = 'New York' OR
-  name = :customer_name AND address.city = 'San Francisco'
-----
+See <<#query>> for examples of valid query syntax.
 
 [#register-view]
-== Registering the View
+=== Register the View
 
-Once you've defined a View, register it with the `AkkaServerless` server by invoking the `addComponent` function.
-
-.index.js
-[source,js,indent=0]
-----
-include::example$js-customer-registry/index.js[tag=register]
-----
+In the View implementation, register the View with Akka Serverless. In addition to passing the service descriptor and a unique identifier, pass any descriptors that define state. In this example, the `customer_domain.proto` descriptor defines the Value Entity state:
 
 .customer-value-entity-view.js
 [source,js,indent=0]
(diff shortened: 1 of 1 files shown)


rstento:
Pinging @mckeeh3, @retgits, @pvlugter for review.

docs/src/modules/javascript/pages/views.adoc
```diff
@@ -4,18 +4,28 @@ include::ROOT:partial$include.adoc[]
 include::partial$views.adoc[]
 
 [#value-entity]
-== View from Value Entity
+== View from a Value Entity
 
-Using an example of a customer registry, you can define a `Customer` in Protobuf as:
+Consider an example of a Customer Registry service with a `customer` Value Entity. When `customer` state changes, the entire state is emitted as a value change. Value changes update any associated Views. To create a View that shows customers by their name:
```
tsonny, line 9: suggestion: "that lists customers by name"

rstento: Done, thanks.

rstento:
@pvlugter, can you please review or suggest a different reviewer? Thanks.

Pinging @retgits.

docs/src/modules/javascript/pages/views.adoc
```diff
@@ -242,4 +222,55 @@ Rebuilding a new View may take some time if there are many events that have to b
 
 The View definitions are stored and validated when a new version is deployed. There will be an error message if the changes are not compatible.
 
+[#query]
+== Query reference
+
+Define View queries in a language that is similar to SQL. The following examples illustrate the syntax for a `customers` entity. To retrieve:
```
retgits, line 228: I think we're missing where the table name in the SQL query comes from (from the name of the table you specify when defining your View service)

retgits:
I think it looks okay (added one comment), but I'd like an engineer to take a look as well.

docs/src/modules/javascript/pages/views.adoc
rstento: I missed that, good catch!

rstento: fixed.

docs/src/modules/javascript/pages/views.adoc
pvlugter: Update not pushed?

docs/src/modules/javascript/pages/views.adoc
```diff
@@ -4,18 +4,28 @@ include::ROOT:partial$include.adoc[]
...
 
-When `Customer` state changes, the entire state is emitted as a value change, which updates any associated Views.
+=== Define the View service descriptor
 
-To get a view of multiple customers by their name, define the View as a `service` in Protobuf:
+To get a view of multiple customers by their name, define the View as a `service` in Protobuf. The descriptor defines:
+
+* How to update the View
+* The source of View data
+* A `table` property that can be any name
```
pvlugter, line 27: It may be useful to say more than just that the table can be given any name. The main point here is that you refer to the given name for the table in queries. So if you have `table: "customers"` then your queries have `SELECT * FROM customers`.

docs/src/modules/javascript/pages/views.adoc
```diff
@@ -38,88 +50,57 @@ include::example$js-customer-registry/customer_view.proto[tag=unary]
...
-  name = :customer_name AND address.city = 'New York' OR
-  name = :customer_name AND address.city = 'San Francisco'
+include::example$js-customer-registry/customer-value-entity-view.js[tag=register]
 ----
 
-[#register-view]
-== Registering the View
-
-Once you've defined a View, register it with the `AkkaServerless` server by invoking the `addComponent` function.
+To use the View in the client, invoke the `addComponent` function. For example:
```
pvlugter, line 68: I don't think this should be `in the client`. It's registering the view as a component in the service / user function (or whatever we're using for this now). `client` is confusing in this context.

pvlugter: The original says:

> Once you've defined a View, register it with the `AkkaServerless` server by invoking the `addComponent` function.

which makes it clear that this is registering in the server or service.

docs/src/modules/javascript/pages/views.adoc
```diff
@@ -38,88 +50,57 @@ include::example$js-customer-registry/customer_view.proto[tag=unary]
...
 
-The following lines in the `.proto` file define a View to consume the `CustomerCreated` and `CustomerNameChanged` events:
+=== Define the View descriptor
+
+A view descriptor:
+
+* Defines update methods for events.
+* Provides the source of the View.
+* Enables transformation updates.
+* Specifies a `table` attribute used by Akka Serverless to store the View. You can use any name.
```
pvlugter, line 100: Again, could be useful to mention that the given table name is used to refer to this table in queries.

pvlugter: Note, it's table `attribute` here, while it was table `property` for value entities. Could be useful to use the same term.

docs/src/modules/javascript/pages/views.adoc
```diff
@@ -147,31 +128,30 @@ include::example$js-customer-registry/customer-event-sourced-view.js[tag=process
...
-.index.js
+.customer-event-sourced-view.js
 [source,js,indent=0]
 ----
-include::example$js-customer-registry/index.js[tag=register-event-sourced]
+include::example$js-customer-registry/customer-event-sourced-view.js[tag=register]
 ----
 
-.customer-event-sourced-view.js
+In the client, use the `addComponent` function to register the View. For example:
```
pvlugter, line 146: `In the client` doesn't make sense here. It's registering with the user function / service / server.

docs/src/modules/javascript/pages/views.adoc
pvlugter: Update not pushed? Yes, the main point to make for the table name is that it's used to refer to the table in queries.

pvlugter (commented):
Looks good. Some comments added.

docs/src/modules/javascript/pages/views.adoc
rstento: hmm, the code is in the client, no, index.js? And, I received a comment that talking about a serverless server was a bit confusing. Changed to: Invoke the `addComponent` function to register the view with the service.

docs/src/modules/javascript/pages/views.adoc
rstento: fixed.

docs/src/modules/javascript/pages/views.adoc
rstento: same change as above.

rstento:
I still need an approving review. Also @pvlugter, my question to reviewers didn't show up because it wasn't in part of the content that got changed. WRT the following statements in the intro:

You might want to retrieve multiple Entities, and retrieve them using an attribute other than the key. Akka Serverless Views allow you achieve this. 

Question: Don't Views also work using Keys? The text implies that Views only work with attributes other than the key. Did it mean to say that you can retrieve multiple entities and you can *also* retrieve them using attributes other than the key?

rstento:
@pvlugter @octonato, I need an approving review, although I'm still glad to make changes if needed.

tsonny (approved):
LGTM

pvlugter:
@rstento, it looks like suggestions were marked resolved, but none of these changes were actually pushed.

docs/src/modules/javascript/pages/views.adoc
pvlugter: Not sure why `index.js` implies a client. Using `service` is good, but has also been seen as ambiguous. Either way, the component is registered via the SDK with the thing that gets deployed as an akka serverless service. A `client` should be reserved for clients that call the deployed service.

docs/src/modules/javascript/pages/views.adoc
pvlugter: fixed where?

docs/src/modules/javascript/pages/views.adoc
pvlugter: Not actually changed in this PR

pvlugter:
> You might want to retrieve multiple Entities, and retrieve them using an attribute other than the key. Akka Serverless Views allow you achieve this.
> 
> Question: Don't Views also work using Keys? The text implies that Views only work with attributes other than the key. Did it mean to say that you can retrieve multiple entities and you can _also_ retrieve them using attributes other than the key?

Yes, probably better as _or_. If you send commands to entities, you can only send to a single entity using its entity key, which is a unique identifier. Views allow indexing on attributes other than the entity key, and the view index key doesn't need to be unique and can return multiple results.

docs/src/modules/javascript/pages/views.adoc
rstento: @pvlugter, I'm not sure what happened here. My commits addressing some of your comments apparently wasn't successful. I'm going to try again.

rstento: See https://github.com/lightbend/akkaserverless-javascript-sdk/pull/139/files line 228.

docs/src/modules/javascript/pages/views.adoc
rstento: See https://github.com/lightbend/akkaserverless-javascript-sdk/pull/139/files

docs/src/modules/javascript/pages/views.adoc
rstento: See https://github.com/lightbend/akkaserverless-javascript-sdk/pull/139/files

docs/src/modules/javascript/pages/views.adoc
rstento: See https://github.com/lightbend/akkaserverless-javascript-sdk/pull/139/files

docs/src/modules/javascript/pages/views.adoc
rstento: See https://github.com/lightbend/akkaserverless-javascript-sdk/pull/139/files

rstento:
Thanks @pvlugter, also pushed this to: https://github.com/lightbend/akkaserverless-javascript-sdk/pull/139/files

Issue #114, opened 2021-07-08
Use consistent proto file naming convention in JS value entities


rstento:
@mckeeh3 commented on [Mon Jun 07 2021](https://github.com/lightbend/akkaserverless-docs/issues/641)

In [Implementing Value Entities in JavaScript](https://docs.akkaserverless.dev/js-services/value-entity.html) change proto file names to either `shopping-cart-api.proto` or `shopping-cart-domain.proto`. 

---

@rstento commented on [Mon Jun 07 2021](https://github.com/lightbend/akkaserverless-docs/issues/641#issuecomment-856185960)

@mckeeh3, this page has already changed (PR in progress) because the code generation tools have changed and this page is bing updated to use the counter Value Entity example from code gen by @andreaTP. Hugh, please take a look at #633, it is now using the convention of <entity-name>-api or <entity-name>-domain.

---

@rstento commented on [Thu Jul 08 2021](https://github.com/lightbend/akkaserverless-docs/issues/641#issuecomment-876601493)

This still isn't consistent, some examples use _ as the separator, others - and the Event Sourced entity page isn't using the service name in the proto file name.

patriknw:
According to the style guide https://developers.google.com/protocol-buffers/docs/style:

> Files should be named lower_snake_case.proto

rstento:
In the samples directory, it looks like all examples are using the style referenced by @patriknw in the previous comment. The page referenced in the initial issue has been fixed. https://docs.akkaserverless.dev/js-services/proto.html still needs to be fixed using the convention and importing from the samples director. However, we don't have a plain shopping cart proto example, we have a value entity and an event sourced entity cart example. The proto overview page could be updated with both, using tabs or we could pick one.

thothmoses:
@rstento Hmm, I'm feeling a bit of an idiot. I know I'm missing something. - I have looked at the page to be fixed ( https://docs.akkaserverless.dev/js-services/proto.html  ) and I don't see anything that needs to be fixed with respect to this issue. The proto files are all referenced using lower_snake_case, and all the source is being pulled via include from the examples - shoppingcart_api.proto and shoppingcart_domain.proto respectively. Please advise.

rstento:
[...]

PR #138, opened 2021-07-22
adding some examples

thothmoses:
Showing the differences between ESE protos and VE protos

A follow-up enhancement from the terminal conversation in Use consistent proto file naming convention in JS value entities #114

### DIFF
--- a/docs/src/modules/javascript/pages/proto.adoc
+++ b/docs/src/modules/javascript/pages/proto.adoc
@@ -13,14 +13,7 @@ Event Sourced Entity::
 --
 [source, proto]
 ----
-// Describes how this domain relates to an event sourced entity
-option (akkaserverless.file).event_sourced_entity = {
-  name: "ShoppingCart"
-  entity_type: "eventsourced-shopping-cart"
-  state: "Cart"
-  events: "ItemAdded"
-  events: "ItemRemoved"
-};
+include::example$js-eventsourced-shopping-cart/proto/shoppingcart_domain.proto[tag=proto_state]
 ----
 --
 Value Entity::
@@ -28,12 +21,7 @@ Value Entity::
 --
 [source, proto]
 ----
-// Describes how this domain relates to a value entity
-option (akkaserverless.file).value_entity = {
-  name: "ShoppingCart"
-  entity_type: "shopping-cart"
-  state: "Cart"
-};
+include::example$js-valueentity-shopping-cart/proto/shoppingcart_domain.proto[tag=proto_state]
 ----
 --
 
--- a/docs/src/modules/javascript/pages/proto.adoc
+++ b/docs/src/modules/javascript/pages/proto.adoc
@@ -5,6 +5,38 @@ include::partial$attributes.adoc[]
 
 Akka Serverless SDKs support `protobuf` descriptors in the https://developers.google.com/protocol-buffers/docs/proto3[Proto3 {tab-icon}, window="new-tab"] Protocol Buffers language. You define command messages, data associated with Entities, and events in `.proto` files. From these definitions, the gRPC compiler creates client and server side code that saves work for you and enables Akka Serverless to serialize message data.
 
+Note that Value Entities and Event Sourced Entities are slightly different based the codegen requirements. Other than the code generation annotations, the proto file are materially equivalent for the same use-case. See the following examples which show the difference.
+
+[.tabset]
+Event Sourced Entity::
++
+--
+[source, proto]
+----
+// Describes how this domain relates to an event sourced entity
+option (akkaserverless.file).event_sourced_entity = {
+  name: "ShoppingCart"
+  entity_type: "eventsourced-shopping-cart"
+  state: "Cart"
+  events: "ItemAdded"
+  events: "ItemRemoved"
+};
+----
+--
+Value Entity::
++
+--
+[source, proto]
+----
+// Describes how this domain relates to a value entity
+option (akkaserverless.file).value_entity = {
+  name: "ShoppingCart"
+  entity_type: "shopping-cart"
+  state: "Cart"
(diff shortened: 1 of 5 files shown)


docs/src/modules/javascript/partials/eventsourced.adoc
```diff
@@ -6,8 +6,6 @@ An Event Sourced Entity must not update its in-memory state directly as a result
 
 When you need to read state in your service, ask yourself _what events should I be listening to_?. When you need to write state, ask yourself _what events should I be emitting_?
```
rstento, line 7: Not sure why the todo is being removed, I don't see an illustration added?

docs/src/modules/javascript/pages/proto.adoc
```diff
@@ -5,6 +5,38 @@ include::partial$attributes.adoc[]
 
 Akka Serverless SDKs support `protobuf` descriptors in the https://developers.google.com/protocol-buffers/docs/proto3[Proto3 {tab-icon}, window="new-tab"] Protocol Buffers language. You define command messages, data associated with Entities, and events in `.proto` files. From these definitions, the gRPC compiler creates client and server side code that saves work for you and enables Akka Serverless to serialize message data.
 
+Note that Value Entities and Event Sourced Entities are slightly different based the codegen requirements. Other than the code generation annotations, the proto file are materially equivalent for the same use-case. See the following examples which show the difference.
```
rstento, line 8: Not sure what "codegen requirements" are? or "code generation annotations"? And, can we pull the code in from the examples directory instead of inline?

rstento (changes requested):
Thanks @thothmoses! I left a few questions.

docs/src/modules/javascript/pages/proto.adoc
thothmoses: Ok, I will tweak that text. I'll have to alter the proto files. The VE is tagged but the section is too large so I'll add tags for a tighter extraction. The ESE is not tagged at all.  I believe tags can be nested within other tags, Correct? We'll see.

docs/src/modules/javascript/partials/eventsourced.adoc
thothmoses: Sorry - the TODO is appearing in the live PROD docs for some reason. It looked correct to me, so this seemed best for now.
https://docs.akkaserverless.dev/java/eventsourced.html

docs/src/modules/javascript/pages/proto.adoc
thothmoses: @pvlugter Hi Peter - I'm trying to pull in:  include::example$js-eventsourced-shopping-cart/proto/shoppingcart_domain.proto[tag=proto_state]
But that sample does not appear to be getting dynamically copied into examples. Could you check this out and fix if possible - or direct me to what I am missing. Thanks.

docs/src/modules/javascript/partials/eventsourced.adoc
rstento: How strange, it does seem to be formatted correctly. Sometimes if you delete it and do it again that fixes it.

thothmoses: ok - I add them back in.

docs/src/modules/javascript/pages/proto.adoc
rstento: @thothmoses, in order to test this, you'll need to follow the procedure I sent in email the other day. The samples are built in the docs/current branch.

thothmoses: @rstento  Thanks, but Nope, that is not the problem all the other samples are being pulled in and are appearing after running Make.
It is just this one file is throwing an error during the build. @pvlugter @patriknw Can you figure this?

pvlugter: @thothmoses, right, that sample is not currently being copied. Only these samples:

https://github.com/lightbend/akkaserverless-javascript-sdk/blob/f7b529a2820700130e0c25bd036345a692b5b6dd/docs/Makefile#L43-L45

You can add the appropriate line to the Makefile:

```
rsync -a --exclude-from=.examplesignore ../samples/js-eventsourced-shopping-cart/ "${managed_examples}/js-eventsourced-shopping-cart/"
```

And check what ends up in `docs/build/src/managed/modules/javascript/examples/` after `make examples`

thothmoses: @pvlugter Perfect, thank you sir. That worked, and I learned something very useful.

docs/src/modules/javascript/pages/proto.adoc
```diff
@@ -5,6 +5,26 @@ include::partial$attributes.adoc[]
 
 Akka Serverless SDKs support `protobuf` descriptors in the https://developers.google.com/protocol-buffers/docs/proto3[Proto3 {tab-icon}, window="new-tab"] Protocol Buffers language. You define command messages, data associated with Entities, and events in `.proto` files. From these definitions, the gRPC compiler creates client and server side code that saves work for you and enables Akka Serverless to serialize message data.
 
+Note that Value Entities and Event Sourced Entities are slightly different based upon codegen requirements. Other than the code generation annotations, the proto file are materially equivalent for the same use-case. See the following examples which show the difference.
```
rstento, line 8:
```suggestion
Note that Value Entities and Event Sourced Entities are slightly different based upon codegen requirements. Other than the code generation annotations, the proto files are materially equivalent for the same use-case. See the following examples which show the difference.
```

rstento (approved):
Sorry, I didn't get back to this sooner! Just one minor typo, but you'll also need to get a dev to review.

PR #140, opened 2021-07-28
Add links 98

### LINKED ISSUE #98: Add links to subsections in Implemention Views
By @mckeeh3 
In Implementing Views in both javaScript and Java, change the three main subsections in the bullet list to links to their associated subsections. I know there are links on the right nav list, but things like these links are a nice touch when re-referencing specific documentation.

Event Sourced Entity events
Value Entities state changes
Messages received from subscribing to topics on a broker
Als0, order the bullets in the same order as the subsections (for us OCD folks).

rstento:
Fixes #98. Also removes partial page which is not needed now that content is not shared between Java and JS.


docs/src/modules/javascript/pages/views.adoc
```diff
@@ -83,7 +98,8 @@ Construct Event Sourced Entity Views from the events that the Entity emits. Buil
 . <<#ES_register>>.
 
 
-The example assumes a `customer_domain.proto` file defines events that will update the View because a name changed:
+The example assumes a `customer_domain.proto` file that defines the events that
+will update the View when  name changes:
```
pvlugter, line 102: Was this going to be `when a name changes`?

pvlugter (approved):
LGTM. We left the partials in case we did move them to a common place, or to update them more easily across repos. Better with the links to sections for this on though.

docs/src/modules/javascript/pages/views.adoc
rstento: Yup, fixed it thanks.

Issue #97, opened 2021-07-07
Need full proto example in "Implementing Views"


rstento:
By @mckeeh3:
In Implementing Views for both JS and Java a full view proto file example is needed. The current examples do not include syntax, package, imports, and view specific option details.

Without these details, it is impossible to create views.

rstento:
Closed via #141

PR #141, opened 2021-07-29
enhancement: show fuller proto files

rstento:
Fixes issue #97. Note, I'm marking this as a draft because it involves both source code and doc and I am not able to build it locally on my machine to check it. 

I also did not add all of the proto file to each example because we are using the same proto file to illustrate all of the different types of views. Instead, I just added the declarations. Are these additions enough so that a user copying them from the doc could create their own views (of each different type)?


pvlugter (approved):
LGTM

PR #142, opened 2021-07-29
Testing views 104

### LINKED ISSUE #104: Add information about testing views
How Views can be tested: https://discuss.lightbend.com/t/how-to-test-views/8427/2

rstento:
Fixes #104

### DIFF
--- a/docs/src/modules/javascript/pages/views.adoc
+++ b/docs/src/modules/javascript/pages/views.adoc
@@ -32,7 +32,7 @@ This example assumes the following `customer` state is defined in a `customer_do
 
 [source,proto,indent=0]
 ----
-include::example$js-customer-registry/customer_domain.proto[tag=domain]
+include::example$js-customer-registry/customer_domain.proto[tags=declarations;domain]
 ----
 
 === Define the View service descriptor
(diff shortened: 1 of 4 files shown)


docs/src/modules/javascript/pages/views.adoc
```diff
@@ -213,6 +214,37 @@ include::example$js-customer-registry/customer_view.proto[tag=wrap-repeated]
...
 <3> The `repeated` field is referenced in the query with `* AS results`.
 
+[#testing]
+== How to test a View
+
+View tests need to create gRPC clients for both the Entity and the View. For example:
+
+[source,javascript]
+----
+include::example$js-customer-registry/integration-test/customer-registry-test.js[lines=17..30]
```
pvlugter, line 224: Would be much better to use tags in the source files, rather than line numbers. We could update the test and never realise it's referenced from the docs.

samples/js-customer-registry/customer_domain.proto
```diff
@@ -12,9 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// tag::declarations[]
```
pvlugter, line 15: Add these tags seem unrelated and unused in this PR.

pvlugter (commented):
Looks fine, but would be good to use tags for code snippets, not line numbers.

samples/js-customer-registry/customer_domain.proto
rstento: oops, slipped in from my other PR. Removed.

pvlugter (approved):
Thanks 👍🏼

PR #145, opened 2021-08-02
doc: add example of multi field entity key

### LINKED ISSUE #144: Add gRPC compound entity keys example
@rstento commented on [Mon Aug 02 2021](https://github.com/lightbend/akkaserverless-java-sdk/issues/184)

@mckeeh3 commented on [Mon Jun 07 2021](https://github.com/lightbend/akkaserverless-docs/issues/637)

In the [Messages](https://docs.akkaserverless.dev/java-services/proto.html#_messages) section of the docs. Please provide a proto file example of a compound multi-field entity key. It is not clear how this is done in proto file syntax.

---

@rstento commented on [Mon Jun 07 2021](https://github.com/lightbend/akkaserverless-docs/issues/637#issuecomment-856284053)

@ennru, could a developer help with this?

---

@ennru commented on [Tue Jun 08 2021](https://github.com/lightbend/akkaserverless-docs/issues/637#issuecomment-856671643)

Thanks for the input. This is not crucial right now.
I believe it is "simply"
```proto
message AddLineItem {
    string user_id = 1 [(akkaserverless.field).entity_key = true];
    string product_id = 2 [(akkaserverless.field).entity_key = true];
```

---

@rstento commented on [Tue Jul 27 2021](https://github.com/lightbend/akkaserverless-docs/issues/637#issuecomment-887719400)

When we are ready to do this, this issue needs to be moved/copied to the Java and the JavaScript SDK repos.

rstento:
Fixes #144

### DIFF
--- a/docs/src/modules/javascript/pages/proto.adoc
+++ b/docs/src/modules/javascript/pages/proto.adoc
@@ -45,13 +45,13 @@ Akka Serverless uses the `entity_key` to know which instance of an Entity a mess
 include::example$js-valueentity-shopping-cart/proto/shoppingcart_api.proto[tag=proto_messages]
 ----
 
-To use a multi-field key, add the `entity_key` notation to each field. For example, if the domain proto file `AddLineItem` element defines both the `user_id` and `product_id` as part of the key, do the same for the service message:
+To use a multi-field key, add the `entity_key` notation to each field. For example, if the domain proto file `SomeMessage` element defines both `first_field` and `second_field` as part of the key, do the same for the service message:
 
 [source,proto]
 ----
-message AddLineItem {
-    string user_id = 1 [(akkaserverless.field).entity_key = true];
-    string product_id = 2 [(akkaserverless.field).entity_key = true];
+message SomeMessage {
+    string first_field = 1 [(akkaserverless.field).entity_key = true];
+    string second_field = 2 [(akkaserverless.field).entity_key = true];
 ----
 
 === Service
@@ -106,13 +106,13 @@ It includes the messages that define the entity state:
 include::example$js-valueentity-shopping-cart/proto/shoppingcart_domain.proto[tag=proto_messages]
 ----
 
-To use a multi-field key, add the `entity_key` notation to each field. For example, if the service API proto file `AddLineItem` element defines both the `user_id` and `product_id` as part of the key, do the same for the service message:
+To use a multi-field key, add the `entity_key` notation to each field. For example, if the service API proto file `SomeMessage` element defines both `first_field` and `second_field` as part of the key, do the same for the service message:
 
 [source,proto]
 ----
-message AddLineItem {
-    string user_id = 1 [(akkaserverless.field).entity_key = true];
-    string product_id = 2 [(akkaserverless.field).entity_key = true];
+message SomeMessage {
+    string first_field = 1 [(akkaserverless.field).entity_key = true];
+    string second_field = 2 [(akkaserverless.field).entity_key = true];
 ----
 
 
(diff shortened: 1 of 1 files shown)


pvlugter (changes requested):
Good to have a compound entity key example, but as mentioned in https://github.com/lightbend/akkaserverless-java-sdk/pull/185, I don't think it should be a shopping cart message as it seems potentially confusing. As we don't have an example with a compound key, maybe just an artificial example for `SomeMessage` would be better.

rstento:
Updated, please re-review.

docs/src/modules/javascript/pages/proto.adoc
```diff
@@ -35,14 +35,25 @@ include::example$js-valueentity-shopping-cart/proto/shoppingcart_api.proto[tag=p
...
+
+Akka Serverless uses the `entity_key` to know which instance of an Entity a message is for. If more than one field is specified as an Entity key, the fields are concatenated together. Akka Serverless serializes Entity keys to strings.
 
 .proto/shoppingcart_api.proto
 [source,proto]
 ----
 include::example$js-valueentity-shopping-cart/proto/shoppingcart_api.proto[tag=proto_messages]
 ----
 
+To use a multi-field key, add the `entity_key` notation to each field. For example, if the domain proto file `SomeMessage` element defines both `first_field` and `second_field` as part of the key, do the same for the service message:
```
pvlugter, line 48: The part about defining entity keys for both domain and service messages doesn't make sense here. Entity keys are only defined on incoming messages for the service, so that messages can be routed to the correct entity.

```suggestion
To use a multi-field key, add the `entity_key` notation to each field. For example, if the `SomeMessage` service message defines both `first_field` and `second_field` as part of the key:
```

docs/src/modules/javascript/pages/proto.adoc
```diff
@@ -35,14 +35,25 @@ include::example$js-valueentity-shopping-cart/proto/shoppingcart_api.proto[tag=p
...
+
+[source,proto]
+----
+message SomeMessage {
+    string first_field = 1 [(akkaserverless.field).entity_key = true];
+    string second_field = 2 [(akkaserverless.field).entity_key = true];
```
pvlugter, line 54: Use 2-space indentation for protobuf:

```suggestion
message SomeMessage {
  string first_field = 1 [(akkaserverless.field).entity_key = true];
  string second_field = 2 [(akkaserverless.field).entity_key = true];
```

docs/src/modules/javascript/pages/proto.adoc
```diff
@@ -95,6 +106,15 @@ It includes the messages that define the entity state:
...
 
+To use a multi-field key, add the `entity_key` notation to each field. For example, if the service API proto file `SomeMessage` element defines both `first_field` and `second_field` as part of the key, do the same for the domain message:
+
+[source,proto]
+----
+message SomeMessage {
+    string first_field = 1 [(akkaserverless.field).entity_key = true];
+    string second_field = 2 [(akkaserverless.field).entity_key = true];
+----
+
```
pvlugter, line 117: Entity keys are not defined on domain messages. This part can be removed.

```suggestion
```

docs/src/modules/javascript/pages/proto.adoc
rstento: Fixed.

docs/src/modules/javascript/pages/proto.adoc
rstento: Done.

pvlugter (approved):

PR #150, opened 2021-08-05
doc: update customer registry sample with ignored event for views

pvlugter:
Refs #108. Update the customer registry sample used for views with a change address command and address changed event, which is ignored by the views.

PR #153, opened 2021-08-06
doc: show how to ignore events for views

### LINKED ISSUE #108: Document use of CatchOthers with Views
From @sean-walsh:
This should be documented for proto and the view impl.

For example, I have a view that only utilizes 2 of many events and the others are ignored. Anything that subscribes to an event sourced entity may need a CatchOthers.

rstento:
Fixes #108 Uses the code provided in #150

### DIFF
--- a/samples/js-customer-registry/customer_view.proto
+++ b/samples/js-customer-registry/customer_view.proto
@@ -140,7 +140,7 @@ service CustomerByNameView {
 
   rpc IgnoreOtherEvents(google.protobuf.Any) returns (domain.CustomerState) {
     option (akkaserverless.method).eventing.in = {
-      event_sourced_entity: "customers"
+      event_sourced_entity: "customers" // <7>
      };
      option (akkaserverless.method).view.update = {
        table: "customers"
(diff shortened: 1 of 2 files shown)


samples/js-customer-registry/customer_view.proto
```diff
@@ -140,7 +140,7 @@ service CustomerByNameView {
 
   rpc IgnoreOtherEvents(google.protobuf.Any) returns (domain.CustomerState) {
```
pvlugter, line 141: Might be better with the callout on the rpc line instead, like it is for <1> and <4>

```suggestion
  rpc IgnoreOtherEvents(google.protobuf.Any) returns (domain.CustomerState) { // <7>
```

samples/js-customer-registry/customer_view.proto
```diff
@@ -140,7 +140,7 @@ service CustomerByNameView {
...
     option (akkaserverless.method).eventing.in = {
-      event_sourced_entity: "customers"
+      event_sourced_entity: "customers" // <7>
```
pvlugter, line 143: Might be better with the callout on the rpc line instead, like it is for <1> and <4>

```suggestion
      event_sourced_entity: "customers"
```

pvlugter (approved):

PR #156, opened 2021-08-12
fix: downgrade to package lockfile version 1 (npm 6)

### LINKED ISSUE #154: Incorrectly generated package-lock.json file in the sdk
### Issue
The current version in master incorrectly assumes that we are using [NPM 7](https://github.com/lightbend/akkaserverless-javascript-sdk/blob/59320f2a314e8de68e75d66f653b28cf6e592d3f/sdk/package-lock.json#L4).

Another issue is that the `packages` [section](https://github.com/lightbend/akkaserverless-javascript-sdk/blob/59320f2a314e8de68e75d66f653b28cf6e592d3f/sdk/package-lock.json#L6) probably shouldn't be part of the versioned `package-lock.json` file, because it's always decided/generated during packaging time.

### Solution
Regenerate the `package-lock.json` with NPM 6, without the `packages` section.

pvlugter:
Resolves #154


package-lock.json
```diff
@@ -1,2180 +1,7 @@
 {
   "name": "akkaserverless-javascript-sdk",
-  "lockfileVersion": 2,
```
janory, line 3: We have changed this everywhere else to `"lockfileVersion": 1`, why are we removing it here?

janory (approved):

PR #166, opened 2021-09-02
feat: Add the TS version of the value entity example

janory:


samples/ts-valueentity-shopping-cart/src/shoppingcart.ts
```diff
@@ -0,0 +1,153 @@
...
+
+
+/**
+ * Handler for add item commands.
+ */
+function addItem(addItem: AddLineItem, cart: State, ctx: Context) {
+  // Validation:
+  // Make sure that it is not possible to add negative quantities
+  if (addItem.quantity < 1) {
+    ctx.fail("Cannot add negative quantity to item " + addItem.productId);
```
pvlugter, line 94: We're moving away from the context-based API, to using what's currently the `Reply`-based API here (and becoming `Effect` in the Java SDK, which includes updates to the state, and will likely be mirrored here as well). Maybe this sample should use `replies` now as the preferred API, and also add typing for the return value of handler functions.

samples/ts-valueentity-shopping-cart/test/testkit.ts
```diff
@@ -0,0 +1,207 @@
...
+ */
+
+/**
+ * Mocks the behaviour of a single Akka Serverless EventSourcedEntity.
+ *
+ * Handles any commands and events, internally tracking the state and maintaining an event log.
+ *
+ * NOTE: Entity IDs are not handled, so all commands are assumed to refer to a single entity.
+ */
+export class MockEventSourcedEntity {
```
pvlugter, line 24: Looks like this `testkit` should be part of the SDK rather than the code generation? Add an issue to follow up on this?

pvlugter (approved):
Nice, LGTM. Could be good to switch to the Reply-based APIs in the samples, as the preferred approach.

pvlugter:
Build failing on the vulnerability check (with `npm audit --production`). Updated in #167.

janory:
> Build failing on the vulnerability check (with `npm audit --production`). Updated in #167.

Thanks for this fix!
Approved, merged, rebased. 👍

samples/ts-valueentity-shopping-cart/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "testSourceDir": "./test",
+    "protoSourceDir": "./proto",
+    "generatedSourceDir": "./lib/generated",
+    "compileDescriptorArgs": []
+  },
+  "scripts": {
+    "start": "node dist/index.js",
+    "test": "ts-mocha test/**/*.test.ts",
+    "integration-test": "ts-mocha integration-test/**/*.integration-test.ts",
+    "build": "akkasls-scripts build && rimraf {src,test,integration-test}/**/*.js && rimraf lib/**/*{shoppingcart,index}.{js,d.ts} && tsc",
```
janory, line 35: @pvlugter the `rimraf` part is a workaround, because the `akkasls-scripts` generates `.js` files, which we have to cleanup on each build. 
Once the codegen becomes TS compatible we can remove this hack.

The cleanup on the `lib` is just small extra, because we don't need those files anymore.

samples/ts-valueentity-shopping-cart/integration-test/shoppingcart.integration-test.ts
```diff
@@ -0,0 +1,101 @@
...
+import { IntegrationTestkit } from "@lightbend/akkaserverless-javascript-sdk";
+import { expect } from "chai"
+import * as proto from "../lib/generated/proto";
+import shoppingcartEnity from '../src/shoppingcart'
+
+type ShoppingCartService = proto.com.example.shoppingcart.ShoppingCartService
+
+type AsyncShoppingCartService = {
+  [K in keyof ShoppingCartService as `${K}Async`]: ShoppingCartService[K]
+}
```
janory, line 26: @pvlugter this is a pretty cool TS feature, we could move this small piece of type inference logic into the `IntegrationTestkit`, when we are working on the https://github.com/lightbend/akkaserverless-javascript-sdk/issues/168 issue. What do you think?

samples/ts-valueentity-shopping-cart/test/testkit.ts
janory: Good idea, issue created: https://github.com/lightbend/akkaserverless-javascript-sdk/issues/168

samples/ts-valueentity-shopping-cart/src/shoppingcart.ts
janory: Updated everywhere.
Can you please check if I am using the `Reply`-based API correctly?
I've never used it before and I was implementing it based on the TCKs examples.
`return {};` (empty response) got replaced by `return replies.noReply();` is that correct?

samples/ts-valueentity-shopping-cart/integration-test/shoppingcart.integration-test.ts
pvlugter: 👍🏼 sounds good

samples/ts-valueentity-shopping-cart/src/shoppingcart.ts
pvlugter: `noReply` should be for no reply at all — different from an empty or default message. It might be translated automatically in this case, but an explicit `replies.message` with an empty is probably correct instead.

Would be good to make `Reply` generic with the message type. The Java SDK is moving on to using `Effect` rather than `Reply`, which I expect may be mirrored here as well. So could be looked at then.

janory: > Would be good to make Reply generic with the message type

What do you mean by this? Could you give a simple example?
Does it worth the effort to change the `Reply` in the SDK if we will migrate to `Effect` soon anyways?

pvlugter: In the Java SDK, replies are generic, `Reply<T>`, where it's the response message type. So a gRPC method that returns an `Empty` message would have a handler that returns `Reply<Empty>`.

Could be worthwhile having the type for `Reply` now, which will just be transferred over to `Effect` if we do that rename too. `Effect` in the Java SDK is an inner class specific to the type of component — actions, value entities, event sourced entities, and so on, have their own different effects — so something similar would be needed.

Issue #163, opened 2021-08-23
TypeScript support for codegen and kickstart


pvlugter:
Support generating TypeScript kickstart projects and code. Probably a switch on the existing scripts to generate TS instead of JS.

The TypeScript code generation could also follow the approach being developed for the Java SDK, where an interface is generated to be implemented, and hooking up with command/event handlers is automatic.

janory:
Don't forget to remove [this](https://github.com/lightbend/akkaserverless-javascript-sdk/pull/166#discussion_r702886921) "cleanup hack" in the context of this issue.

janory:
@pvlugter I will start working this now.

janory:
Some work was already done related to this task.
It can be found [here](https://github.com/lightbend/akkaserverless-javascript-sdk/tree/163-typeScript_support_for_codegen/codegen/js-gen/src/main/scala/com/lightbend/akkasls/codegen/ts).

PR #170, opened 2021-09-07
feat: add TS version for the rest of the samples

janory:
Implements this: https://github.com/lightbend/akkaserverless-javascript-sdk/issues/163

## Changes:
- added the TS version for the rest of the samples
- accordingly to the comment from the previous PR replaced all the `replies.noReply()` with `replies.message({})`
- added prettier and linter for the TS implementations
- moved the `TS` and `JS` samples into different sub folders for the better transparency


samples/ts/ts-customer-registry/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "ts-customer-registry",
+  "version": "0.0.0",
+  "description": "Akka Serverless JavaScript SDK sample application for customer registry",
```
pvlugter, line 4: Shall we call it the `TypeScript SDK` or have it mentioned as a TypeScript sample in some way?

```suggestion
  "description": "Akka Serverless TypeScript SDK sample application for customer registry",
```

pvlugter (commented):
Ok, great to have more converted samples!

The main reason is for the docs, so I think we only really need the samples that are referenced from the docs. So that should be the shopping cart for event sourced entities, the shopping cart and counter for value entities, and the customer registry for views.

Some things needed given the changes (and failing build):

- docs: the Makefile needs the paths updated (and snippet references if the sample paths in the docs change)
- sample tests and integration tests in CI build need updating given the new locations
- TS samples should be added to the CI build as well to be tested

And then including the TS versions of the snippets in the docs (with tabbed code examples).

janory:
> Ok, great to have more converted samples!
> 
> The main reason is for the docs, so I think we only really need the samples that are referenced from the docs. So that should be the shopping cart for event sourced entities, the shopping cart and counter for value entities, and the customer registry for views.

You are right. I was not sure which ones are used in the docs, so I've migrated all of them. 
But I think it's good for our users anyways, that they have all the samples in TS as well.

> 
> Some things needed given the changes (and failing build):
> 
> * docs: the Makefile needs the paths updated (and snippet references if the sample paths in the docs change)
> * sample tests and integration tests in CI build need updating given the new locations
> * TS samples should be added to the CI build as well to be tested

Thanks for the pointers. I've tried to fix all the mentioned things, please double check it.

> And then including the TS versions of the snippets in the docs (with tabbed code examples).

This PR is already quite big, would it be okay, if we would do this in a separate PR once this one war merged?

pvlugter (approved):
LGTM. Let's merge and then get the tabbed snippets in the docs.

Issue #174, opened 2021-09-14
Include TypeScript snippets alongside the JavaScript snippets in the documentations


janory:

beritou:
This is a great idea. What do you imagine it would look like?

I like how the Akka docs have separate tabs for Scala & Java. Not sure how much effort this would be in Akka Serverless docs.

<img width="1075" alt="image" src="https://user-images.githubusercontent.com/6062111/133310601-5c574afe-bedb-41ed-aebd-a9d0752f0401.png">

https://doc.akka.io/docs/akka/current/typed/actors.html#first-example

retgits:
That's indeed what we're planning :)

janory:
@beritou This is exactly how I've implemented it. There will be two separate tabs van for JS and one for TS.

beritou:
Excellent

PR #175, opened 2021-09-14
feat: add TS snippets to the docs

janory:
Resolves https://github.com/lightbend/akkaserverless-javascript-sdk/issues/174.

### DIFF
--- a/docs/src/modules/javascript/examples/action/src/action.ts
+++ b/docs/src/modules/javascript/examples/action/src/action.ts
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Action, replies } from "@lightbend/akkaserverless-javascript-sdk";
+import * as proto from "../lib/generated/proto";
+
+type Request = proto.akkaserverless.tck.model.action.Request;
+type ProcessGroup = proto.akkaserverless.tck.model.action.ProcessGroup;
+
+const { Response, ProcessGroup } = proto.akkaserverless.tck.model.action;
+
+export const tckModel: Action = new Action(
+    "proto/action.proto",
+    "akkaserverless.tck.model.action.ActionTckModel"
+).setCommandHandlers({
+  ProcessUnary: processUnary,
+  ProcessStreamedIn: processStreamedIn,
+  ProcessStreamedOut: processStreamedOut,
+  ProcessStreamed: processStreamed
+});
+
+function processUnary(request: Request): replies.Reply {
+  return createReplyForGroup(ProcessGroup.create(request.groups[0]));
+}
+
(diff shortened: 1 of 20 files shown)


retgits (approved):
LGTM!

docs/src/modules/javascript/examples/test/eventsourced/shoppingcart.ts
```diff
@@ -0,0 +1,199 @@
...
+
+const entity: EventSourcedEntity = new EventSourcedEntity(
+    ["shoppingcart.proto", "domain.proto"],
+    "example.shoppingcart.ShoppingCartService",
+    {
+        persistenceId: "shopping-cart",
+        snapshotEvery: 100
+    }
+);
+// end::entity-class[]
```
pvlugter, line 32: Examples shouldn't be copied here manually. Use the compiled and tested samples, which are copied into the examples automatically (the rsync in the Makefile).

docs/src/modules/javascript/pages/serialization.adoc
```diff
@@ -26,3 +28,21 @@ const entity = new EventSourcedEntity(
...
+const entity: EventSourcedEntity = new EventSourcedEntity(
+  ["shoppingcart.proto", "domain.proto"],
+  "com.example.shoppingcart.ShoppingCart",
+  {
+    persistenceId: "shopping-cart",
+    snapshotEvery: 5, // Usually you wouldn't snapshot this frequently, but this helps to demonstrate snapshotting
+    includeDirs: ["./"],
+    serializeFallbackToJson: true // Enables JSON support for persistence
+  }
+);
```
pvlugter, line 47: Let's use the compiled and tested sample, rather than inline.

docs/src/modules/javascript/examples/action/src/action.ts
```diff
@@ -0,0 +1,111 @@
...
+
+import { Action, replies } from "@lightbend/akkaserverless-javascript-sdk";
+import * as proto from "../lib/generated/proto";
+
+type Request = proto.akkaserverless.tck.model.action.Request;
+type ProcessGroup = proto.akkaserverless.tck.model.action.ProcessGroup;
+
+const { Response, ProcessGroup } = proto.akkaserverless.tck.model.action;
+
+export const tckModel: Action = new Action(
```
pvlugter, line 25: The action example is using the TCK? This doesn't look that good as an example.

pvlugter (commented):
👍🏼 Looks good for a TS version of the examples... but also looks like the JS examples have an outdated approach from when these were in the main docs.

All the example snippets should come from the samples, which are compiled and tested, so we can maintain them more easily and they don't go out-of-date.

There's no proper action example? Using the TCK model as an example doesn't read well.

Realise these are existing issues with the JS docs. If we don't care, then fine to merge as is.

retgits:
@pvlugter 

> Realise these are existing issues with the JS docs. If we don't care, then fine to merge as is.

Totally understand that, I'd go with "let's merge now and make progressive updates after that". We need a bunch of updates in the doc anyway :)

docs/src/modules/javascript/examples/test/eventsourced/shoppingcart.ts
janory: Yes, I was quite confused about the fact that some of the examples are coming from a "static" place, but some of them from the dynamically copied samples.

Since I was not sure why is it made like this, I've replicated the same behaviour of the JS examples.
So where the JS examples took the static path, I did the same for the TS ones.

docs/src/modules/javascript/examples/action/src/action.ts
janory: Yes, it doesn't correspond to the `proto` which is shown above on the page.
I was thinking about what to do with this one. 
First I wanted to skip it entirely for TS, because adding a TS version based on the `proto` would have look strange when switching between the tabs. 

But then I've just decided to add the typed version of the TCK example to keep it consistent.

PR #184, opened 2021-09-16
chore(docs): add bundles for quickstarts

retgits:
This PR:

- Adds bundling script and makefile target to create quickstart zip files (copied from the Java SDK)
- Adds antora tags to the JavaScript Event Sourced Shopping cart so it can be used in the quickstarts

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -52,8 +52,8 @@ examples:
 	rsync -a --exclude-from=.examplesignore ../samples/ts/ts-eventsourced-shopping-cart/ "${managed_examples}/ts-eventsourced-shopping-cart/"
 
 bundles:
-	bin/bundle.sh --zip "${managed_attachments}/js-customer-registry-quickstart.zip" ../samples/js/js-customer-registry-quickstart
-	bin/bundle.sh --zip "${managed_attachments}/ts-customer-registry-quickstart.zip" ../samples/ts/ts-customer-registry-quickstart
+	bin/bundle.sh --zip "${managed_attachments}/js-customer-registry-quickstart.zip" ../samples/js/js-customer-registry
+	bin/bundle.sh --zip "${managed_attachments}/ts-customer-registry-quickstart.zip" ../samples/ts/ts-customer-registry
 	bin/bundle.sh --zip "${managed_attachments}/js-eventsourced-shopping-cart.zip" ../samples/js/js-eventsourced-shopping-cart
 	bin/bundle.sh --zip "${managed_attachments}/ts-eventsourced-shopping-cart.zip" ../samples/ts/ts-eventsourced-shopping-cart
 
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -22,7 +22,7 @@ build: dev
 clean:
 	rm -rf build
 
-managed: attributes apidocs examples
+managed: attributes apidocs examples bundles
 	mkdir -p "${src_managed}"
 	cp src/antora.yml "${src_managed}/antora.yml"
 
@@ -51,6 +51,11 @@ examples:
 	rsync -a --exclude-from=.examplesignore ../samples/ts/ts-valueentity-counter/ "${managed_examples}/ts-valueentity-counter/"
 	rsync -a --exclude-from=.examplesignore ../samples/ts/ts-eventsourced-shopping-cart/ "${managed_examples}/ts-eventsourced-shopping-cart/"
 
+bundles:
+	bin/bundle.sh --zip "${managed_attachments}/js-customer-registry-quickstart.zip" ../samples/js/js-customer-registry-quickstart
+	bin/bundle.sh --zip "${managed_attachments}/ts-customer-registry-quickstart.zip" ../samples/ts/ts-customer-registry-quickstart
+	bin/bundle.sh --zip "${managed_attachments}/js-eventsourced-shopping-cart.zip" ../samples/js/js-eventsourced-shopping-cart
+	bin/bundle.sh --zip "${managed_attachments}/ts-eventsourced-shopping-cart.zip" ../samples/ts/ts-eventsourced-shopping-cart
 
 dev: clean managed validate-xrefs dev-html
 
--- a/docs/bin/bundle.sh
+++ b/docs/bin/bundle.sh
@@ -27,6 +27,7 @@ function _remove_doc_tags {
   # note: use commands that are compatible with both GNU sed and BSD (macOS) sed
   find "$dir" -type f -exec sed -i.bak "/tag::[^\[]*\[.*\]/d" {} \; -exec rm -f {}.bak \;
   find "$dir" -type f -exec sed -i.bak "/end::[^\[]*\[.*\]/d" {} \; -exec rm -f {}.bak \;
+  find "$dir" -type f -exec sed -i.bak "s/\/\/\s<.>//g" {} \; -exec rm -f {}.bak \;
 }
 
 function _bundle {
@@ -54,8 +55,8 @@ function _bundle {
 
   _remove_doc_tags "$sample_bundle_dir"
 
-  pushd "$bundle_dir" > /dev/null
-  zip -q -r "$zip_file" "$sample_name"
+  pushd "$sample_bundle_dir" > /dev/null
+  zip -q -r "$zip_file" .
   popd > /dev/null
 
   echo "Bundled $sample as $zip"
--- a/docs/bin/bundle.sh
+++ b/docs/bin/bundle.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+#
+# Bundle a sample project for download.
+#
+# Usage:
+#   bundle.sh --zip <file> <dir>
+
+set -euo pipefail
+
+function _script_path {
+  local source="${BASH_SOURCE[0]}"
+  while [ -h "$source" ] ; do
+    local linked="$(readlink "$source")"
+    local dir="$(cd -P $(dirname "$source") && cd -P $(dirname "$linked") && pwd)"
+    source="$dir/$(basename "$linked")"
+  done
+  echo ${source}
+}
+
+readonly script_path=$(_script_path)
+readonly script_dir="$(cd -P "$(dirname "$script_path")" && pwd)"
+readonly docs_dir="$(cd "$script_dir/.." && pwd)"
+readonly bundle_dir="$docs_dir/build/bundle"
+
+function _remove_doc_tags {
+  local -r dir="$1"
+  # note: use commands that are compatible with both GNU sed and BSD (macOS) sed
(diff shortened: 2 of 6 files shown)


docs/Makefile
```diff
@@ -51,6 +51,11 @@ examples:
 	rsync -a --exclude-from=.examplesignore ../samples/ts/ts-valueentity-counter/ "${managed_examples}/ts-valueentity-counter/"
 	rsync -a --exclude-from=.examplesignore ../samples/ts/ts-eventsourced-shopping-cart/ "${managed_examples}/ts-eventsourced-shopping-cart/"
 
+bundles:
+	bin/bundle.sh --zip "${managed_attachments}/js-customer-registry-quickstart.zip" ../samples/js/js-customer-registry-quickstart
```
patriknw, line 55: Note the issue about nested dir https://github.com/lightbend/akkaserverless-java-sdk/issues/380

retgits: Do we want to fix that first before merging this in?

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
```diff
@@ -13,15 +13,28 @@
 // limitations under the License.
 
 // This is the public API offered by the shopping cart entity.
-
+/* tag::declarations[] */
```
pvlugter, line 16: I thought asciidoc required the tags to be at the end of the line. May be good to check that this works, compared with:

```proto
// tag::declarations[]
```

docs/bin/bundle.sh
```diff
@@ -0,0 +1,64 @@
...
+readonly script_path=$(_script_path)
+readonly script_dir="$(cd -P "$(dirname "$script_path")" && pwd)"
+readonly docs_dir="$(cd "$script_dir/.." && pwd)"
+readonly bundle_dir="$docs_dir/build/bundle"
+
+function _remove_doc_tags {
+  local -r dir="$1"
+  # note: use commands that are compatible with both GNU sed and BSD (macOS) sed
+  find "$dir" -type f -exec sed -i.bak "/tag::[^\[]*\[.*\]/d" {} \; -exec rm -f {}.bak \;
+  find "$dir" -type f -exec sed -i.bak "/end::[^\[]*\[.*\]/d" {} \; -exec rm -f {}.bak \;
```
pvlugter, line 29: Are there any callouts in the shopping cart samples being bundled? If there are, then we need some sed to remove those too.

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
retgits: I've copied these tags verbatim from the current doc sources (https://github.com/lightbend/akkaserverless-docs/blob/master/docs/modules/src/examples/js-shopping-cart/shoppingcart_api.proto) to make sure that I used the same tags I had before (which made it easier for me to update the quickstart 😅). After rendering, it comes out as https://developer.lightbend.com/docs/akka-serverless/quickstart/cr-value-entity-javascript.html. With this PR we can remove the final code snippets from the docs and make sure all samples in the docs come from the SDK repos.

docs/bin/bundle.sh
retgits: I'm not sure I know what you mean with callouts?

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
pvlugter: Ok, cool

docs/bin/bundle.sh
pvlugter: https://docs.asciidoctor.org/asciidoc/latest/verbatim/callouts/

Any code blocks that have the numbered callouts with `// <1>` or similar

pvlugter: There are some in the customer registry samples

docs/Makefile
retgits: Updated with the same changes as @pvlugter did for Java

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
retgits: Just to keep the style the same, I updates the tags to `// tag`

docs/bin/bundle.sh
retgits: Ah thanks :) I've added a regex to do that.

docs/bin/bundle.sh
```diff
@@ -0,0 +1,65 @@
...
+  find "$dir" -type f -exec sed -i.bak "s/\/\/\s<.>//g" {} \; -exec rm -f {}.bak \;
```
pvlugter, line 30: I don't think this is compatible across both BSD sed and GNU sed.

I don't think `\s` is supported across both (even with extended enabled). Could use `[[:space:]]` but I think we can just expect a literal space and not worry about tab characters. Also, could be useful to match digits specifically, and could allow for more than one (just in case an example gets to 10 callouts). And while we're here, could be useful to strip the leading whitespace as well.

So sticking to basic regex, maybe something like this instead: `"s/ *\/\/ *<[0-9][0-9]*>//g"`

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
```diff
@@ -56,22 +60,21 @@ service ShoppingCartService {
 
   rpc AddItem(AddLineItem) returns (google.protobuf.Empty) {
     option (google.api.http) = {
-      post: "/cart/{cart_id}/items/add"
-      body: "*"
+        post: "/cart/{user_id}/items/add"
+        body: "*"
```
pvlugter, line 64: Should be 2-space indentation, as it was before.

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
```diff
@@ -56,22 +60,21 @@ service ShoppingCartService {
 
   rpc AddItem(AddLineItem) returns (google.protobuf.Empty) {
     option (google.api.http) = {
-      post: "/cart/{cart_id}/items/add"
-      body: "*"
+        post: "/cart/{user_id}/items/add"
```
pvlugter, line 63: To change to `{user_id}` the field in `AddLineItem` would also need to be changed. This should have an error currently, as it won't find the field.

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
```diff
@@ -56,22 +60,21 @@ service ShoppingCartService {
...
+        body: "*"
     };
   }
 
   rpc RemoveItem(RemoveLineItem) returns (google.protobuf.Empty) {
-    option (google.api.http).post = "/cart/{cart_id}/items/{product_id}/remove";
+      option (google.api.http) = {
+          post: "/cart/{user_id}/items/{product_id}/remove";
+      };
```
pvlugter, line 71: 2-space indents

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
```diff
@@ -56,22 +60,21 @@ service ShoppingCartService {
...
-    option (google.api.http) = {
-      get: "/carts/{cart_id}"
-      additional_bindings: {
-        get: "/carts/{cart_id}/items"
-        response_body: "items"
-      }
-    };
+      option (google.api.http) = {
+          get: "/carts/{user_id}"
+      };
```
pvlugter, line 77: 2-space indents

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
```diff
@@ -56,22 +60,21 @@ service ShoppingCartService {
...
-      body: "*"
+        post: "/cart/{user_id}/items/add"
+        body: "*"
     };
   }
 
   rpc RemoveItem(RemoveLineItem) returns (google.protobuf.Empty) {
-    option (google.api.http).post = "/cart/{cart_id}/items/{product_id}/remove";
+      option (google.api.http) = {
+          post: "/cart/{user_id}/items/{product_id}/remove";
```
pvlugter, line 70: The semicolon needs to be removed here. Failed in the CI build when compiling the protobuf.

pvlugter:
```
Error: illegal name ';' (/home/circleci/project/samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto, line 70)
```

docs/bin/bundle.sh
retgits: Updated to match the new regex you suggested

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
retgits: That was a copy/paste error 😅

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
retgits: Thanks! I've removed it

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
```diff
@@ -62,16 +66,15 @@ service ShoppingCartService {
   }
 
   rpc RemoveItem(RemoveLineItem) returns (google.protobuf.Empty) {
-    option (google.api.http).post = "/cart/{cart_id}/items/{product_id}/remove";
+      option (google.api.http) = {
+        post: "/cart/{cart_id}/items/{product_id}/remove"
+      };
```
pvlugter, line 71:
```suggestion
    option (google.api.http) = {
      post: "/cart/{cart_id}/items/{product_id}/remove"
    };
```

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
```diff
@@ -62,16 +66,15 @@ service ShoppingCartService {
...
-    option (google.api.http) = {
-      get: "/carts/{cart_id}"
-      additional_bindings: {
-        get: "/carts/{cart_id}/items"
-        response_body: "items"
-      }
-    };
+      option (google.api.http) = {
+        get: "/carts/{cart_id}"
+      };
```
pvlugter, line 77:
```suggestion
    option (google.api.http) = {
      get: "/carts/{cart_id}"
    };
```

pvlugter:
Hmm, build is failing on unzipping protoc.

pvlugter (approved):

PR #186, opened 2021-09-17
Document inter service calls

johanandren:
Refs https://github.com/lightbend/akkaserverless-docs/issues/903


samples/js/js-doc-snippets/src/delegatingservice.js
```diff
@@ -0,0 +1,53 @@
...
+
+// FIXME sample waited for client to be ready?
+
+action.commandHandlers = {
+  async AddAndReturn(request, ctx) {
+    const empty = await counterClient.increase({counter_id: request.counter_id, value: 1});
+    const currentCounter = await counterClient.getCurrentCounter({counter_id: request.counter_id});
+    return replies.message({ value: currentCounter.value });
+  }
+};
```
johanandren, line 50: I haven't seen this actually working yet (stuck on some js sdk protoc compilation issue I don't understand)

Got a hint from Andrea about this sample on how to create a client which I based this on: 
https://github.com/badsyntax/grpc-js-typescript/blob/master/examples/grpc-proto-loader/client.ts

johanandren: Works now. 👍

samples/js/js-doc-snippets/src/delegatingservice.js
```diff
@@ -0,0 +1,92 @@
...
+  // but actually calling without callback fails on param validation in grpc-js
+  // so for now we do this promise dance ourselves
+  async AddAndReturn(request, ctx) {
+    const increaseDone = await new Promise((resolve, reject) => {
+      counterClient.increase({counterId: request.counterId, value: 1}, // <5>
+        (error, emptyResult) => {
+          if (error) reject(error);
+          else resolve(emptyResult);
+        });
+    });
```
johanandren, line 77: I think this is the only annoyance right now, the apidoc of the service makes it look as though it can be called without any callbacks and that it will return a promise in that case, but when actually trying that out, some internal parameter validation in proto-js barfs on no callbacks passed. :/

docs/src/modules/javascript/pages/call-another-service.adoc
```diff
@@ -0,0 +1,54 @@
+= Calling other services
+
+== Akka Serverless services
+
+In some cases it is useful to call a component in another service, for example interacting with multiple entities or actions, aggregating their return values or performing a multi-step workflow. xref:forwarding.adoc[Forwards] and side-effects allow us to trigger other services but does not make it possible to compose and transform the results of calling them.
```
ennru, line 5:
```suggestion
In some cases it is useful to call a component in another service, for example interacting with multiple entities or actions, aggregating their return values or performing a multi-step workflow. xref:forwarding.adoc[Forwards] and side-effects allow us to trigger other services but can't compose or transform the results.
```

docs/src/modules/javascript/pages/call-another-service.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+Calling other Akka Serverless services in the same project from an Action is done by invoking them over gRPC much like how an external client would. The service is however identified only by the name it has been deployed as, Akka Serverless takes care of routing requests to the service and keeping the data safe by encrypting the connection for us.
+
+In this sample we will make an action that does two sequential calls to the xref:value-entity.adoc[Value Entity Counter] service, deployed with the service name "counter".
```
ennru, line 9:
```suggestion
In this sample we will show an action that does two sequential calls to the xref:value-entity.adoc[Value Entity Counter] service, deployed with the service name "counter".
```

docs/src/modules/javascript/pages/call-another-service.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+This is how the counter descriptor looks with all annotations removed:
+
+[source,protobuf,indent=0]
+.proto/com/example/counter_api.proto
+----
+include::example$js-doc-snippets/proto/com/example/counter_api.proto[tag=other-service-api]
+----
+
+The proto-js will now generate a client for the service in `lib/generated/proto.js` when we compile the project.
```
ennru, line 25:
```suggestion
The proto-js tool will now generate a client for the service in `lib/generated/proto.js` when we compile the project.
```

ennru (commented):
Some wording suggestions.

samples/js/js-doc-snippets/src/delegatingservice.js
pvlugter: The integration testkit also creates clients. When adding integration tests for the samples, I also added a promisify for the client methods (originally https://github.com/lightbend/akkaserverless-framework/pull/617) so that each method has an `Async` version that returns a promise.

https://github.com/lightbend/akkaserverless-javascript-sdk/blob/227123339a17192404fd2f5b1341bb024e3e604d/sdk/src/integration-testkit.ts#L110-L124

We could do something similar here. Or rather than patching could provide our own client wrapper if we want to only have the Promise returning methods (without `Async).

johanandren: Ah, nice, so we could provide something like in the Java sdk to look up the client and return one with async-patched methods. I like that idea.

johanandren:
Alright, this should be ready for final review now!

Not the most beautiful API with the promisify, but nicer than manually doing promises at least (was hoping to monkey patch out of the box but didn't figure out how to do that, since it needs to do it to the service client constructor)

samples/js/js-doc-snippets/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "type": "module",
+  "engines": {
+    "node": ">=14.0.0",
+    "npm": ">=6.0.0"
+  },
+  "dependencies": {
+    "@lightbend/akkaserverless-javascript-sdk": "0.30.0"
+  },
+  "devDependencies": {
+    "@lightbend/akkasls-scripts": "0.30.0",
```
pvlugter, line 13: These could also be latest version `0.33.1`. There should probably be an update all the samples after release, like for Java SDK.

samples/js/js-doc-snippets/proto/com/example/counter_api.proto
```diff
@@ -0,0 +1,37 @@
...
+syntax = "proto3";
+
+package com.example;
+
+import "google/protobuf/empty.proto";
+
+message IncreaseValue {
+    string counter_id = 1;
+    int32 value = 2;
+}
```
pvlugter, line 12: Use 2-space indents?

pvlugter (approved):
LGTM

> Not the most beautiful API with the promisify, but nicer than manually doing promises at least (was hoping to monkey patch out of the box but didn't figure out how to do that, since it needs to do it to the service client constructor)

Certainly looks better than manually resolving promises. The integration testkit creates clients for all the registered components. For this, it could probably accept a constructor and create the client for you. Something that can be followed up on anyway.

Should have changes from #189 applied to the new sample.

samples/js/js-doc-snippets/package.json
pvlugter: made it the same as other samples `0.33.0` for now

pvlugter:
Have applied the small review suggestions directly: updated for the local development changes from #189, reindented proto with 2 spaces, bumped to the same SDK version as other samples.

pvlugter:
@johanandren I've started hacking together something for promisifying clients automatically in #193

PR #187, opened 2021-09-21
docs: add documentation for replicated entities

### LINKED ISSUE #146: Add Replicated Entity page back to JavaScript SDK doc pages
This task is to bring back the old documentation on replicated entities, but the content still needs to be updated to reflect the new implementation.

pvlugter:
Resolves #146.

Add docs for replicated entities, with replicated counter map shopping cart samples for JavaScript and TypeScript. These docs don't yet go as far as the Java SDK docs for replicated entities, which have an example for every replicated data type.

This only documents replicated entities as they are now. There's no codegen for replicated entity, so that's not covered. And typing for replicated data could be improved (making replicated entity and replicated data types generic) but currently just documenting as is.


pvlugter:
Needs changes from #189 applied to the new samples.

PR #213, opened 2021-10-15
chore: label issues and PRs

ennru:
Try to add labels whenever issues or PRs are created.﻿

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,16 @@
+# https://github.com/marketplace/actions/labeler
+
+as-framework:
+  - '**/*'
+
+javascript-sdk:
+  - sdk/**/*.js
+  - codegen/js-gen/**/*
+  - samples/js/**/*
+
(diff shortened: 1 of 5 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,16 @@
...
+as-framework:
+  - '**/*'
+
+javascript-sdk:
+  - sdk/**/*.js
+  - codegen/js-gen/**/*
+  - samples/js/**/*
+
+typescript-sdk:
+  - sdk/**/*.ts
```
pvlugter, line 12: It's not that simple to distinguish, as some of the core code is in TS files. The TypeScript SDK is the JavaScript SDK plus type definitions, which are currently defined as jsdoc, either in the JS files or in the the .jsdoc files when the source is in TS.

pvlugter (approved):
LGTM, but the JS/TS distinction is not clear cut.

PR #215, opened 2021-11-02
Exclude more files from quickstart bundles

patriknw:
* exclude all hidden such as .idea .DS_Store and such

Refs lightbend/akkaserverless-docs#1043

### DIFF
--- a/samples/js/js-customer-registry/.bundleignore
+++ b/samples/js/js-customer-registry/.bundleignore
@@ -1,3 +1,4 @@
 node_modules
 package-lock.json
 user-function.desc
+.*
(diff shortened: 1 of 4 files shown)


samples/js/js-customer-registry/.bundleignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 package-lock.json
 user-function.desc
+.*
```
patriknw, line 4: @pvlugter is it ok to exclude a such `.*` files/directories for Javascript?

pvlugter: Yes, I think it's fine to exclude all the dot files. If we did want to include a `.gitignore` then we can un-exclude it again. But seems good to me.

pvlugter (approved):

PR #221, opened 2021-12-02
feat: added jwt claim metadata support

jroper:
This depends on #220 being merged first.

This adds JWT claims metadata support.


johanandren (approved):
LGTM

PR #224, opened 2022-01-03
chore: upgrade proxy to 0.8.5

octonato:
We need this in order to use the new annotations in the Javascript SDK.

Draft because on top of #226.

Ready for review otherwise.


janory (approved):
LGTM!

PR #226, opened 2022-01-03
chore: fix version of @types/node to 14.18.3

octonato:
Build in main start to fail because of silent update of grpc-js library. This only affecting TS compilation. 

```node
node_modules/@grpc/grpc-js/build/src/object-stream.d.ts:11:18 - error TS2430: Interface 'IntermediateObjectWritable<T>' incorrectly extends interface 'Writable'.
  The types returned by 'end(...)' are incompatible between these types.
    Type 'void' is not assignable to type 'this'.
      'this' could be instantiated with an arbitrary type which could be unrelated to 'void'.
```

Without this change, the grpc-js lib is bumped to 1.4.5 causing the error above.

### DIFF
--- a/samples/ts/ts-customer-registry/package-lock.json
+++ b/samples/ts/ts-customer-registry/package-lock.json
@@ -343,9 +343,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "16.9.6",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-16.9.6.tgz",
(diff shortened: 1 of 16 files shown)


octonato:
I changed this PR. The problem is related to the version of @types/node, but it manifest itself when compiling grpc-js (in our case). 

More info here https://github.com/grpc/grpc-node/issues/2002

My latest commit pins the version used in the sdk to 14.18.3 and then, on each TS sample, we explict set the same version. Without that, the sample uses version 17.0.7 which also causes the same issue.

johanandren (approved):

PR #230, opened 2022-01-04
feat: new codegen annotations

octonato:
Draft because on top of #229 and also not working it. Summiting PR to confirm that build is still green.


octonato:
Ready for review.

We are now capable of building the intermediate AST, but we still need to take into consideration the case for names Actions and Views. I plan to do it on follow-up PR. We will need to add more tests. 

We can also decide to reduce the scope for now. What we have now has feature parity with the previous annotation and therefore we can decide to leave named Actions and Views and unnamed Entities out.

codegen/core/src/test/resources/test-files/codegen-annotation/generate-from-samples.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
+set -euo pipefail
+IFS=$'\n\t'
+
+# Re-generates sample proto file descriptors from the source
+# Requires https://github.com/lightbend/akkaserverless-java-sdk to be cloned locally, and its path provided
```
johanandren, line 6: Why is the Java SDK needed, couldn't it just invoke npm build for each of the samples and then pick up the updated proto desc insteads?

johanandren (approved):
LGTM, but weird to depend on local fork of Java SDK just to get access to proxy protocol/compile descriptors.

codegen/core/src/test/resources/test-files/codegen-annotation/generate-from-samples.sh
octonato: This is wrong, I will remove it. Also from the Java SDK. 

At one point, we added the new annotations in the SDK and then I needed it. I forgot to remove this when I put the protos back in the framework.

octonato: Done in e49069b

PR #233, opened 2022-01-05
feat: use new codegen annotations in seed templates

octonato:
WIP, the initial goal is to check how the npm-js-tests will behave.


ennru:
This needs to be re-done for the Kalix seed template.

PR #236, opened 2022-01-07
chore: suppress warning

octonato:


codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
```diff
@@ -211,6 +211,7 @@ object ModelBuilder {
     }
   }
 
+  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
```
johanandren, line 214: So we are using wart remover in this build for some reason, weird since we never use that anywhere else AFAIK.

johanandren (approved):

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
octonato: Yes, this is from the time codegen an apart project. We also usually use ScalaTest and not MUnit and we didn't change it either.

PR #250, opened 2022-03-02
fix: use "static-module" instead of "static" only target

janory:
References https://github.com/lightbend/akkaserverless/issues/6192

```
 -t, --target     Specifies the target format. Also accepts a path to require a custom target.

                   json          JSON representation
                   json-module   JSON representation as a module
                   proto2        Protocol Buffers, Version 2
                   proto3        Protocol Buffers, Version 3
                   static        Static code without reflection (non-functional on its own)
                   static-module Static code without reflection as a module
```

Source: https://www.npmjs.com/package/protobufjs#pbjs-for-javascript

### DIFF
--- a/npm-js/akkasls-scripts/bin/akkasls-scripts.js
+++ b/npm-js/akkasls-scripts/bin/akkasls-scripts.js
@@ -61,7 +61,7 @@ const scriptHandlers = {
     runOrFail(
       'Building static Javascript definitions from proto',
       path.resolve('node_modules', '.bin', 'pbjs'),
-      [...protoFiles, '-t', 'static', '-o', protoJs],
+      [...protoFiles, '-t', 'static-module', '-o', protoJs],
       { shell: true },
     );
 


johanandren (approved):
I have no clue, but I trust you. LGTM

pvlugter:
I have no idea either, and haven't looked more closely at `akkasls-scripts`. Looks like this static generation was just for generating the type definitions, and not expected to be used directly? I do remember that the SDK expects the reflection-based one — I think it needs the type descriptor or something, which was only provided there.

janory:
> Looks like this static generation was just for generating the type definitions, and not expected to be used directly?

Yeah, but I don't understand why would we do this limitation, probably it was not intentional.

> I do remember that the SDK expects the reflection-based one — I think it needs the type descriptor or something, which was only provided there.

Yes, if the user's logic wants to emit stuff for the context only the reflection-based solution work.
Although if it's about returning something to the client the statically generated one can be used as well.
Check out my sample app: 
For the initial state, reflection:
https://github.com/janory/chirper/blob/d5a3c66c9dae6ee79bb6ca97f547d76849756e98/service/src/auth-entity/index.ts#L84
https://github.com/janory/chirper/blob/d5a3c66c9dae6ee79bb6ca97f547d76849756e98/service/src/auth-entity/index.ts#L91

For the user facing api, the statically generated ones:
https://github.com/janory/chirper/blob/d5a3c66c9dae6ee79bb6ca97f547d76849756e98/service/src/auth-entity/index.ts#L49

pvlugter:
> Although if it's about returning something to the client, the statically generated one can be used as well.

Makes sense, using it as an object to map to the protobuf message.

PR #253, opened 2022-03-09
feat: allow specifying a custom status when failing

jroper:
This allows setting a custom status when failing, by adding an optional status parameter to the `ctx.fail()` and `replies.failure()` methods.


pvlugter (approved):
Cool, looks good. CI is failing on prettier check.

PR #279, opened 2022-04-01
Proxy 1.0.0-M2 and new type_url for primitives

johanandren:

PR #280, opened 2022-04-01
Update js-valueentity-shopping-cart version to 0.0.4

raboof:
0.0.3 is still on the old protocol, might be nice to
have them separate

### DIFF
--- a/samples/js/js-valueentity-shopping-cart/package.json
+++ b/samples/js/js-valueentity-shopping-cart/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-valueentity-shopping-cart",
-  "version": "0.0.3",
+  "version": "0.0.4",
   "type": "module",
   "engines": {
     "node": ">=13.0.0",


samples/js/js-valueentity-shopping-cart/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "js-valueentity-shopping-cart",
-  "version": "0.0.3",
+  "version": "0.0.4",
```
patriknw, line 3: below is `kalix-javascript-sdk": "0.33.5"`, shouldn't we update that? and I would like to publish a new version that includes #279

johanandren: There's a script that updates all sample dependencies like this _after_ doing a release, (they are tested with current as part of PR validation however)

raboof: > below is kalix-javascript-sdk": "0.33.5", shouldn't we update that?

ow, hadn't noticed that :/ - yes, definitely

> and I would like to publish a new version that includes https://github.com/lightbend/akkaserverless-javascript-sdk/pull/279

:+1:

patriknw: I'm not sure I understand. I'm fine with merging this and then cut a new release. Note that we have already published 1.0.0-M3 and next would be M4

johanandren (approved):

samples/js/js-valueentity-shopping-cart/package.json
raboof: ah, gotcha, I didn't realize 0.33.5 is from before the renaming (because the artifact name *was* updated in the package.json here, confusing). No need to wait then :+1:

johanandren: Bumping samples in https://github.com/lightbend/akkaserverless-javascript-sdk/pull/282 (but that isn't blocking a release if you want to make one @patriknw )

PR #282, opened 2022-04-01
Bump sample deps to 1.0.0-M4

johanandren:


samples/js/js-doc-snippets/package.json
```diff
@@ -7,10 +7,10 @@
     "npm": ">=6.0.0"
   },
   "dependencies": {
-    "@kalix-io/kalix-javascript-sdk": "0.33.5"
+    "@kalix-io/kalix-javascript-sdk": "1.0.0-M3"
```
patriknw, line 10: does this M3 already include #279 ? Otherwise we should release a M4 first

johanandren: I don't think it is super important to keep updating these with the milestones, the important part will be to bump once 1.0.0 is released. It might be good to get off 0.33 at least, where the samples would be completely broken because the protocol and APIs changed, but we can definitely wait for M4

raboof (approved):

Issue #284, opened 2022-04-05
Status code reply APIs not available from typescript


johanandren:
Not sure exactly how that works but I think there needs to be something in the replies.jsdoc, the additions in replies.ts is not enough to end up in the compiled index.d.ts

johanandren:
More specifically, this signature: https://github.com/lightbend/akkaserverless-javascript-sdk/blob/main/sdk/src/reply.ts#L233
Does not show up in the compiled ts, I think it can be added through https://github.com/lightbend/akkaserverless-javascript-sdk/blob/main/sdk/src/reply.jsdoc but I don't quite understand how.

PR #286, opened 2022-04-05
fix: Status code was not accessible from typescript #284

johanandren:
Will also need a forward port to the rename-branch

Note that in addition to making the `GrpcStatus` enum public in TS this also moved it from `akkaserverless.replies.GrpcStatus` directly to `akkaserverless.GrpcStatus`

### DIFF
--- a/sdk/src/akkaserverless.jsdoc
+++ b/sdk/src/akkaserverless.jsdoc
@@ -99,3 +99,28 @@
  *                             distributed to remaining replicas in the background.
  * @property {number} ALL Updates will be written immediately to all replicas.
  */
+
+/**
+ * The GRPC status codes.
+ *
+ * @readonly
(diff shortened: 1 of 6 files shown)


sdk/src/akkaserverless.jsdoc
```diff
@@ -99,3 +99,28 @@
  *                             distributed to remaining replicas in the background.
  * @property {number} ALL Updates will be written immediately to all replicas.
  */
+
+/**
```
johanandren, line 103: Needs to be 1:1 duplicated i the jsdoc or else it is no available from the compiled/generated TS API.

janory (approved):
LGTM! 👍

pvlugter (approved):
👍🏼 Yes, this is pretty awkward. Would be good to finish the conversion to TypeScript.

PR #288, opened 2022-04-06
Added additional types for Typescript

jroper:
This continues work from #286, adding Typescript type info for other features that have been added.


johanandren (approved):
LGTM, good catch.

PR #292, opened 2022-04-07
docs: new view features

pvlugter:
Refs https://github.com/lightbend/kalix-proxy/issues/1335. Same docs as https://github.com/lightbend/akkaserverless-java-sdk/pull/897. Only for `kalix-main`.


johanandren:
A bit annoying that we have these in both docs, perhaps the query language reference should go in the central docs once instead?

pvlugter:
> A bit annoying that we have these in both SDK docs, perhaps the query language reference should go in the central docs once instead?

Yeah. Or maybe a docs module in the proxy. It's good when code changes and doc changes are together, and the query syntax is part of the proxy. Query snippets could also be tested.

pvlugter:
There could be other docs that are best paired with the proxy codebase as well.

Issue #293, opened 2022-04-08
Convert codebase to TypeScript


pvlugter:
The codebase has been partially converted over to TypeScript. This can continue in steps. Having everything in TypeScript should make maintenance easier, type declarations can be generated directly (rather than using jsdoc), API docs can be generated using tsdoc, and it lays the groundwork for further improvements.

- [x] convert codebase to TypeScript
- [x] generate type declarations using tsc, remove jsdoc
- [x] generate API docs using tsdoc

PR #294, opened 2022-04-08
chore: convert protobuf support to TypeScript

pvlugter:
Refs #293. Breaking up into smaller steps. Only targeting `kalix-main` for this.

Convert the protobuf any support and tests to TypeScript. Direct translation, just add typing.


johanandren (approved):
LGTM

PR #297, opened 2022-04-14
chore: convert action to TypeScript

pvlugter:
Refs #293. Convert action and action-support to TypeScript.

Trying to be a direct and faithful translation of what's there. Not sure the test coverage would pick up on everything.

With the way that the contexts for unary and streamed calls are created and used, have forced expectations with some casts and non-null assertions.

There was also a `cancel` on the streaming in, but there is no cancel on these server-side interfaces, so I've commented that out for now. The `cancel` method should be removed from the streaming-in context, if cancelling is only client-side.


sdk/src/action-support.ts
```diff
@@ -557,53 +599,56 @@ class ActionHandler {
...
-     */
-    this.ctx.cancel = () => {
-      if (this.call.cancelled) {
+    ctx.cancel = () => {
+      if (call.cancelled) {
         this.streamDebug('cancel invoked when already cancelled.');
       } else {
-        this.call.cancel();
+        // FIXME: there doesn't seem to be any server-side cancel, only client-side, so what was this doing before?
+        // call.cancel();
```
johanandren, line 615: Hmmm. Tried looking into the grpc-js interfaces but I can't find anything suitable-cancel-like either.

johanandren (approved):

PR #299, opened 2022-04-20
fix: more clear error that a view will be stuck when update method missing

johanandren:


pvlugter (approved):
And rebased for CI and the kalix-main/main changes.

PR #308, opened 2022-04-27
rename: some remaining akkaserverless to kalix renaming

pvlugter:
Refs https://github.com/lightbend/kalix-proxy/issues/1234

Noticed that there were still a few mentions of Akka Serverless. Also a couple of small fixes on the codegen side.  

And the test proto descriptors used for codegen were also outdated, using akkaserverless annotations — not sure why the tests were still successful. Haven't looked closer at this, but have re-generated these, as they came up in searching for akkaserverless.


codegen/js-gen/src/main/scala/io/kalix/codegen/js/EntityServiceSourceGenerator.scala
```diff
@@ -99,7 +99,7 @@ object EntityServiceSourceGenerator {
     val typedefPath =
       sourceDirectory.toAbsolutePath
         .relativize(generatedSourceDirectory.toAbsolutePath)
-        .resolve(service.fqn.name.toLowerCase())
+        .resolve(entity.fqn.name.toLowerCase())
```
pvlugter, line 102: small fix, the generated files use the entity name not the service name, typedef imports were wrong

patriknw (approved):
LGTM

PR #310, opened 2022-04-28
chore: convert event sourced entity to TypeScript

pvlugter:
Refs #293. Draft, based on #309.


johanandren (approved):

pvlugter:
Also applied the entity id check here, as for value entity.

PR #318, opened 2022-05-06
feat: adding file to show kalix SDK version

franciscolopezsancho:
References https://github.com/lightbend/kalix-proxy/issues/786


docs/src/modules/javascript/attachments/latest-version.txt
```diff
@@ -0,0 +1 @@
+1.0.0-M5
```
raboof, line 1: Should we refer to 'milestone' versions publically? If we want people to use it externally shouldn't we call it 1.0.1?

raboof (approved):

docs/src/modules/javascript/attachments/latest-version.txt
franciscolopezsancho: I don't have a strong opinion about it. I'm thinking in the context of `kalix-proxy` retrieving the version of the SDK used in the user function for later comparing it to this `latest-version`. Do you think that it will be alright to do your suggested conversion in the proxy as well?
On the other hand, I was thinking that as we publish this version with milestone already [here](https://www.npmjs.com/package/@kalix-io/kalix-javascript-sdk) would make sense for a user to see the same version in the docs. Again I'm far from having a strong opinion about this.

franciscolopezsancho: You are very right. I checked on the other references to the version in the project and after seeing that the samples were hardcoded I wrongly assume it was also the case here. But now that you mention I found docs/bin/version.sh. I'll add this to the make file.

pvlugter: Yes, let's have this automatically generated instead. Also in the JVM SDK:https://github.com/lightbend/kalix-jvm-sdk/pull/952#issuecomment-1120500658

Issue #326, opened 2022-05-11
Support serialization of generated protobuf


pvlugter:
Currently serialized state requires the reflective protobuf message, which includes the `$type` for creating the type url. This makes it awkward from TypeScript, where the reflective and statically generated types need to be mixed for state types.

Protobuf.js does now generate a `getTypeUrl` message (https://github.com/protobufjs/protobuf.js/pull/1463).

This hasn't been released yet (no release for a year), although it looks work is being done to get a new version released. We could otherwise release from a fork temporarily, published under `@kalix-io`, to get this functionality.

pvlugter:
New version of protobufjs has been published, so using only the statically generated protobuf types can be supported now.

Issue #320, opened 2022-05-09
Generic types and Serializable


pvlugter:
The `Serializable` type is currently a union that includes `any`, making the whole type `any` so it's untyped. The `any` should probably be `object` instead, as indicated by the original jsdoc type (which converts `Object` to `any`).

Generics could be used for classes, especially those that have state that can extend `Serializable`: entities and replicated data types. Type parameters can have defaults, so the previous behaviour can also be supported in the same way.

pvlugter:
Still more to do: entities, handlers, contexts.

pvlugter:
This will also affect codegen, which should use the typed versions instead of the generated workaround that it has currently. Probably worthwhile updating some codegen things at the same time, such as using `Reply`. Could also affect the generated `testkit.js`, which should really be part of the SDK directly.

PR #327, opened 2022-05-11
feat: make replicated data types generic

pvlugter:
Refs #320.

Make replicated data types generic. Replicated entity state is not yet generic, so this is just the underlying step towards that. Entities may need a bit of reworking to have this ripple through the contexts in particular.

The typing for serialized state is still awkward in TypeScript, with it requiring the reflective Message type, but should be able to improve on this and use the generated protobuf directly (see #326).

The type parameters will default to `Serializable` to have the same behaviour as before. Some tests are updated, others left as is.

Note that this used single letter type parameters at first (`T`, `E`, `K`, `V` and so on), but the latest TypeScript docs are now using full names, so I've followed the full name convention here as well.

### DIFF
--- a/sdk/src/serializable.ts
+++ b/sdk/src/serializable.ts
@@ -35,7 +35,7 @@ export interface TypedJson {
 export type Serializable =
   | protobuf.Message
   | TypedJson
-  | any
+  | object
   | string
   | number
   | boolean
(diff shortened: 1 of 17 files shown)


sdk/src/serializable.ts
```diff
@@ -35,7 +35,7 @@ export interface TypedJson {
 export type Serializable =
   | protobuf.Message
   | TypedJson
-  | any
+  | object
```
pvlugter, line 38: This change means that `Serializable` is no longer always just `any`. The handlers and callbacks can't use the Serialiable union type directly now, as it would require the implementation to accept all thse types. Entities and their handlers and contexts will also become generic, with a parameter that extends Serializable instead.

johanandren (approved):

PR #269, opened 2022-03-29
rename: switch to kalix protocol (snapshot for now)

johanandren:
CI integration tests won't pass because no proxy image available yet

PR #329, opened 2022-05-15
fix: set component settings for actions

pvlugter:
Forward headers for actions was broken. Looks like this line was inadvertently removed in #269 with the renaming.

Must be missing tests for this. As it's part of the SDK protocol, probably useful to have it in the TCK.


pvlugter:
Since it's just restoring a line, I'll merge without review and trigger a release with the fix. Have checked in manually.

Issue #336, opened 2022-05-23
Rename packages?


pvlugter:
When the SDK was published under `@lightbend` it made sense to qualify with `akkaserverless`. Now that we're using the `@kalix-io` organisation for publishing, it could be better to simplify things, aligning with how npm packages are usually named. And do this before 1.0 final. These names are always fully qualified, and used in imports as well, so the organisation already states that it's `kalix`. 

Suggestions:

- rename `@kalix-io/kalix-javascript-sdk` -> `@kalix-io/sdk`
- rename `@kalix-io/kalix-scripts` -> `@kalix-io/scripts`
  - keep the binary as `kalix-scripts`
- use `@kalix-io/testkit` as a separate dev dependency package
  - #80
- maybe rename the template: `@kalix-io/create-kalix-entity` -> `@kalix-io/create-service`
  - if service is the term that will continue to be used

pvlugter:
I'll update this in a draft PR so we can see what it would be like.

pvlugter:
It's a shame we couldn't use `@kalix` organisation, that would be pretty clean as `@kalix/sdk`, `@kalix/testkit`, `@kalix/scripts` for the modules.

PR #339, opened 2022-05-25
feat: add testkit package

### LINKED ISSUE #80: Make testkit part of SDK instead of generated from codegen
The javascript codegen generates a testkit for value entities and event sourced entities that are very similar to Akka's EventSourcedBehaviorTestkit. 

Instead of generating this over and over, we should make it part of the SDK itself.

pvlugter:
Resolves #80.

Move the testkit from codegen, and the integration testkit in the SDK, into a `@kalix-io/testkit` package.

Separate package so that it can be a dev dependency only. Uses the naming suggested in #336.

This is a breaking change at the package level — the `@kalix-io/testkit` dependency needs to be added to projects, the generated testkit removed, imports updated. Otherwise it's the same testkit support and behaviour as before.

Uses the typed mock variations (from the TypeScript samples, already updated for generic components), so there's full support for TypeScript.

The testkit package has a peer dependency on the SDK for `^1.0.0-M7`, which is the minimum version with the generic components (#338). Once 1.0.0 final is released, this could be updated to be `^1`, to just require a semver compatible SDK. Usually the SDK and testkit dependencies will simply have the same version in a project anyway.

The samples are using testkit `1.0.0-M6`, which doesn't exist, but the CI builds will install the locally built version. This will be corrected to an actual published version later, when samples are updated post-release.

Will check that the CI setup is correct and working.


testkit/src/mock-event-sourced-entity.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  export type CommandNames<Entity> = keyof CommandHandlersType<Entity>;
+  export type EventNames<Entity> = keyof EventHandlersType<Entity>;
+
+  /**
+   * Mocks the behaviour of the command context object within Kalix.
+   *
+   * By default, calls to {@link MockEventSourcedEntity.handleCommand} will
+   * construct their own instance of this class, however for making assertions on
+   * forwarding or emmitted effects you may provide your own.
```
johanandren, line 140:
```suggestion
   * forwarding or emitted effects you may provide your own.
```

johanandren (approved):
LGTM, much better than a generated testkit in each project imo.

testkit/src/mock-event-sourced-entity.ts
pvlugter: 👍🏼 didn't read the old doc comments

PR #341, opened 2022-05-25
rename: simplify package names

### LINKED ISSUE #336: Rename packages?

pvlugter:
Resolves #336. Draft on top of #339.

Try out simplified package names.

- `@kalix-io/kalix-javascript-sdk` -> `@kalix-io/sdk`
- `@kalix-io/kalix-scripts` -> `@kalix-io/scripts`

Plus `@kalix-io/create-kalix-entity` -> `@kalix-io/create-kalix-service`

Also remove `npm-js` and `NPMJS` directories and references. Move modules:

- `/npm-js/kalix-scripts/` -> `/scripts/`
- `/npm-js/create-kalix-entity/` -> `/create/`

The samples refer to non-existing packages, should have local versions installed in the CI build. Will work when published (and bundled quickstarts are updated).

Noticed that there are completely out-of-date integration tests in codegen. Will maybe look at those separately, or simply remove them.

CI should catch any issues.


pvlugter:
CI failing. There's a bootstrap problem, of course, for the other modules that depend on sdk, as we do an install of published dependencies before installing the locally built sdk.

Will need to change the testing temporarily, or maybe some manual intervention, if we want to go ahead with this.

pvlugter:
Rebased. Ready for opinions on whether to go ahead with this rename...

pvlugter:
Since the testkit package hasn't been released yet, maybe its name should be aligned with the others: `@kalix-io/testkit` -> `@kalix-io/kalix-testkit`.

PR #349, opened 2022-06-01
fix: testkit mock command handling on failures

### LINKED ISSUE #347: testkit: command handling tries to serialize/deserialize undefined messages
The mock entities have `handleCommand` methods that run the reply message through serialization:

https://github.com/lightbend/kalix-javascript-sdk/blob/0782b7af0c7d9ac25246e8f2d0a4f11c3dad02da/testkit/src/mock-value-entity.ts#L93-L95

But they do this regardless of whether the message is defined — and it will be undefined for failures. Update to something like:

```ts
return message
  ? grpcMethod.responseDeserialize(grpcMethod.responseSerialize(message))
  : undefined;
```

pvlugter:
Resolves #347

Make the fix for when the message is undefined on failures. Add tests for the mock entities to test this.

Make the test example services shared between tests and integration tests. Codegen and kalix-scripts is a bit awkward to use here still, as it doesn't yet support TypeScript and makes some other assumptions — so generate the type declarations manually and commit these files too. Entities in the JS SDK also support async command handlers, which the codegen doesn't generate for currently. Mock command handling will also need to be async. Will make those changes separately.


johanandren (approved):
LGTM!

PR #350, opened 2022-06-02
fix: support async command handlers in codegen and testkit

pvlugter:
Entity command handlers can already be async. Add support for this to the codegen and testkit.

Testkit hasn't been published yet, so not exactly breaking, but `handleCommand` in the mock entities is now async. If we wanted to make it compatible with what was there before, we could also have two versions of mocked command handling, where the sync version throws if the command handler returns a Promise. But simpler to just have one method, explicitly async.


johanandren (approved):
LGTM

Issue #354, opened 2022-06-07
Support ACLs and self-identity


johanandren:
We need to add support, samples and or docs for:

 * Providing a proto descriptor with a default ACL for an entire JS/TS Kalix Service (In the JVM SDK a `kalix_proxy.proto` is found and picked up if present) and specific docs for how that is done (compare with https://github.com/lightbend/kalix-jvm-sdk/pull/981 )
 * Pick up self identity token and pass along as a request header in cross-component requests (see https://github.com/lightbend/kalix-jvm-sdk/pull/988 )
 * Pick up proxy host name discovery and pass to self-grpc-clients?

johanandren:
#366 should potentially be done as well before this can be considered done.

PR #355, opened 2022-06-08
Acl support

johanandren:
References #354 

 * pass along proxy hostname and auth headers to grpc clients and inject 
    - **breaking API change**: clients can no longer be created before `preStart` has been called triggered by discovery
 * APIs for inspecting principal
 * default acl for an entire service (docs only)
 * testkit support for fake principal
 * test coverage
 * figure out why self-calls does not work getting `14 UNAVAILABLE: No connection established` - test container needs to inform SDK about proxy port

### DIFF
--- a/docs/src/modules/javascript/pages/access-control.adoc
+++ b/docs/src/modules/javascript/pages/access-control.adoc
@@ -35,10 +35,8 @@ In the unit test Mock components, the ACLs are ignored.
 
 == ACLs when running integration tests
 
-FIXME not implemented yet
+When running integration tests, ACLs are enabled by default but can be explicitly disabled per test by running the test with `IntegrationTestKitOptions` setting `aclCheckingEnabled` to `true`.
 
-When running integration tests, ACLs are enabled by default but can be explicitly disabled per test by running the test with `Settings.withAclDisabled`.
+For integration tests that call other services that have ACLs limiting access to specific service names `IntegrationTestKitOptions.withServiceName` allows specifying what the service identifies itself as to other services.
 
-For integration tests that call other services that have ACLs limiting access to specific service names `Settings.withServiceName` allows specifying what the service identifies itself as to other services.
-
-`KalixTestKit.getGrpcClientForPrincipal` makes it possible to get a integration test client that is authenticated with specific credentials for calling a service with ACLs.
+`IntegrationTestkit.clientsForPrincipal` makes it possible to get a integration test client that is authenticated with specific credentials for calling a service with ACLs.
--- a/docs/src/modules/javascript/pages/access-control.adoc
+++ b/docs/src/modules/javascript/pages/access-control.adoc
@@ -24,7 +24,6 @@ This is the default ACL included in the quickstarts and project templates, it al
 
 == Inspecting the principal inside a service
 
-FIXME not implemented yet
 
 Checking the ACLs are in general done for you by Kalix, however in some cases programmatic access to the principal of a call can be useful.
 
@@ -32,7 +31,7 @@ Accessing the principal of a call inside of a service is possible through the re
 
 == ACLs when running unit tests
 
-In the generated unit test testkits, the ACLs are ignored.
+In the unit test Mock components, the ACLs are ignored.
 
 == ACLs when running integration tests
 
--- a/docs/src/modules/javascript/pages/access-control.adoc
+++ b/docs/src/modules/javascript/pages/access-control.adoc
@@ -1,0 +1,45 @@
+= Access Control Lists (ACLs)
+
+This section describes functionality in the JavaScript/TypeScript SDK related to ACLs (Access Control Lists), for information about how the ACLs are defined and control access see https://docs.kalix.io/services/using-acls.html[Using ACLs]
+
+== Default ACL in project templates
+
+If no ACLs is defined at all in a Kalix service Kalix will allow requests from both other services and the internet to all components of a Kalix service.
+
+The Kalix quickstarts include a less permissive ACL for the entire service, to not accidentally make services available to the public internet, just like the one described in the next section.
+
+== Defining an ACL for the entire Kalix Service
+
+A default ACL for the entire Kalix Service can be defined by placing a `kalix_policy.proto` file among the protobuf descriptors of the service, and including it in your build. It should only contain a `kalix.file(acl)` annotation:
+
+[source,proto,indent=0]
+.proto/com/example/kalix_policy.proto
+----
+include::example$js-doc-snippets/proto/com/example/kalix_policy.proto[tag=default]
+----
+<1> Import the needed Kalix annotations from `kalix/annotations.proto`
+<2> Allow access from all other services, but not the public internet
+
+This is the default ACL included in the quickstarts and project templates, it allows calls from any other Kalix service deployed in the same project, but denies access from the internet.
+
+== Inspecting the principal inside a service
+
+FIXME not implemented yet
+
+Checking the ACLs are in general done for you by Kalix, however in some cases programmatic access to the principal of a call can be useful.
+
+Accessing the principal of a call inside of a service is possible through the request metadata `Metadata.principals()`. The `Metadata` for a call is available through the context (`actionContext`, `commandContext`) of the component.
+
+== ACLs when running unit tests
+
+In the generated unit test testkits, the ACLs are ignored.
+
+== ACLs when running integration tests
+
+FIXME not implemented yet
+
+When running integration tests, ACLs are enabled by default but can be explicitly disabled per test by running the test with `Settings.withAclDisabled`.
+
+For integration tests that call other services that have ACLs limiting access to specific service names `Settings.withServiceName` allows specifying what the service identifies itself as to other services.
+
+`KalixTestKit.getGrpcClientForPrincipal` makes it possible to get a integration test client that is authenticated with specific credentials for calling a service with ACLs.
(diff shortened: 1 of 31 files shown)


sdk/src/action.ts
```diff
@@ -90,11 +95,21 @@ export class Action<
...
   }
 
+  preStart(settings: PreStartSettings): void {
+    this.clients = GrpcUtil.clientCreators(
+      this.root,
+      this.grpc,
+      settings.proxyHostname,
+      settings.proxyPort,
+      settings.identificationInfo,
+    );
```
johanandren, line 110: Not setting this until discovery breaks the current samples which creates a client immediately on start.

johanandren: Still went with this breaking change for now, couldn't come up with anything sensible to allow create early but later injection. Perhaps we can improve in the future.

johanandren:
That was one marathon PR 😅 

Ready for review now!

johanandren:
Manually tested a service with ACLs, and self-calls in dev/prod and also using the new principal metadata API. Some final fixes but now it works.

docs/src/modules/javascript/pages/access-control.adoc
```diff
@@ -0,0 +1,42 @@
+= Access Control Lists (ACLs)
+
+This section describes functionality in the JavaScript/TypeScript SDK related to ACLs (Access Control Lists), for information about how the ACLs are defined and control access see https://docs.kalix.io/services/using-acls.html[Using ACLs]
```
pvlugter, line 3:
```suggestion
This section describes functionality in the JavaScript/TypeScript SDK related to ACLs (Access Control Lists), for information about how the ACLs are defined and control access see https://docs.kalix.io/services/using-acls.html[Using ACLs].
```

docs/src/modules/javascript/pages/access-control.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+== Default ACL in project templates
+
+If no ACLs is defined at all in a Kalix service Kalix will allow requests from both other services and the internet to all components of a Kalix service.
+
+The Kalix quickstarts include a less permissive ACL for the entire service, to not accidentally make services available to the public internet, just like the one described in the next section.
+
+== Defining an ACL for the entire Kalix Service
+
+A default ACL for the entire Kalix Service can be defined by placing a `kalix_policy.proto` file among the protobuf descriptors of the service, and including it in your build. It should only contain a `kalix.file(acl)` annotation:
```
pvlugter, line 13: Should that be `(kalix.file).acl` annotation, rather than `kalix.file(acl)`?

```suggestion
A default ACL for the entire Kalix Service can be defined by placing a `kalix_policy.proto` file among the protobuf descriptors of the service, and including it in your build. It should only contain a `(kalix.file).acl` annotation:
```

docs/src/modules/javascript/pages/access-control.adoc
```diff
@@ -0,0 +1,42 @@
...
+----
+<1> Import the needed Kalix annotations from `kalix/annotations.proto`
+<2> Allow access from all other services, but not the public internet
+
+This is the default ACL included in the quickstarts and project templates, it allows calls from any other Kalix service deployed in the same project, but denies access from the internet.
+
+== Inspecting the principal inside a service
+
+
+Checking the ACLs are in general done for you by Kalix, however in some cases programmatic access to the principal of a call can be useful.
```
pvlugter, line 28: maybe a bit smoother as 

```suggestion
Checking the ACLs is generally done for you by Kalix, however in some cases programmatic access to the principal of a call can be useful.
```

docs/src/modules/javascript/pages/access-control.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+Accessing the principal of a call inside of a service is possible through the request metadata `Metadata.principals()`. The `Metadata` for a call is available through the context (`actionContext`, `commandContext`) of the component.
+
+== ACLs when running unit tests
+
+In the unit test Mock components, the ACLs are ignored.
+
+== ACLs when running integration tests
+
+When running integration tests, ACLs are enabled by default but can be explicitly disabled per test by running the test with `IntegrationTestKitOptions` setting `aclCheckingEnabled` to `true`.
```
pvlugter, line 38: Should be disabled by setting to `false`?

```suggestion
When running integration tests, ACLs are enabled by default but can be explicitly disabled per test by running the test with `IntegrationTestKitOptions` setting `aclCheckingEnabled` to `false`.
```

docs/src/modules/javascript/pages/access-control.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+For integration tests that call other services that have ACLs limiting access to specific service names `IntegrationTestKitOptions.withServiceName` allows specifying what the service identifies itself as to other services.
+
+`IntegrationTestkit.clientsForPrincipal` makes it possible to get a integration test client that is authenticated with specific credentials for calling a service with ACLs.
```
pvlugter, line 42:
```suggestion
`IntegrationTestkit.clientsForPrincipal` makes it possible to get an integration test client that is authenticated with specific credentials for calling a service with ACLs.
```

docs/src/modules/javascript/pages/call-another-service.adoc
```diff
@@ -24,19 +24,21 @@ include::example$js-doc-snippets/proto/com/example/counter_api.proto[tag=other-s
...
 ----
 include::example$js-doc-snippets/src/delegatingservice.js[tag=delegating-action]
 ----
 <1> Include the `proto` file of the other service in the action service descriptors.
 <2> Create the client using the provided creators to get async methods. Defaults to a cleartext connection as TLS is handled transparently for us.
-<3> Use the name that the other Kalix service was deployed as, in this case `counter` and port `80`.
+<3> Use the name that the other Kalix service was deployed as, in this case `counter`.
 <4> We can now call the various methods on the other service with their method names directly on the client. The calls return promises so we can use `await/async` to interact with them.
 
+**Note:** The client can only be created once the service has been started by Kalix, for example triggered by a call, not from the constructor.
```
pvlugter, line 40: 👍🏼 and use asciidoc `NOTE:`?

```suggestion
NOTE: The client can only be created once the service has been started by Kalix, for example triggered by a call, not from the constructor.
```

sdk/src/action.ts
pvlugter: > Still went with this breaking change for now, couldn't come up with anything sensible to allow create early but later injection. Perhaps we can improve in the future.

Makes sense. The components are created directly. If it was a class-based approach for defining the component behaviour, then could be easier? But that would be a bigger breaking change again.

sdk/src/kalix.ts
```diff
@@ -237,17 +243,34 @@ export interface Component {
    * @internal
    */
   register(allComponents: ServiceMap): ComponentServices;
+
+  /**
+   * Internal method for setting up the component.
+   */
+  preStart(settings: PreStartSettings): void;
```
pvlugter, line 250: Add an `@internal` annotation to mark properly as internal, and it will also be excluded from the packaged type declarations.

```suggestion
  /**
   * Internal method for setting up the component.
   *
   * @internal
   */
  preStart(settings: PreStartSettings): void;
```

testkit/example/src/action-with-acl.ts
```diff
@@ -0,0 +1,94 @@
...
+        input.field +
+        ', principals: ' +
+        context.metadata.principals().get().map(principalToString),
+    };
+  },
+  DelegateToSelf: async (input) => {
+    // FIXME TS lookup doesn't seem to quite work for nested packages, needs this mess:
+    const clientFactory = (
+      (action.clients!.com as GrpcClientLookup).example as GrpcClientLookup
+    ).ExampleActionWithACLService as GrpcClientCreator;
```
pvlugter, line 83: Yeah, this is messy. The `proto-loader-gen-types` could possibly be used for users too, to create the right structure and client creators.

testkit/src/integration-testkit.ts
```diff
@@ -45,32 +49,60 @@ export namespace IntegrationTestkit {
...
+    }
+
+    this.options.delayDiscoveryUntilProxyPortSet = true;
+
+    // sensible defaults for missing values
+    if (!this.options.dockerImage) {
+      this.options.dockerImage = defaultDockerImage;
+    }
+    if (!this.options.serviceName) {
+      this.options.serviceName = 'self';
```
pvlugter, line 101: There's `nameOfService` in `IntegrationTestKitOptions` and then `options.serviceName` here. Are they meant to be the same option?

pvlugter (commented):
LGTM, from what I can tell from just a read through.

sdk/src/action.ts
johanandren: One thing we could perhaps do is to pass the clients via the various `context`s instead of as a field on the component object itself, that way it is obvious when it is available only when there is a context.

testkit/example/src/action-with-acl.ts
johanandren: Tracked by separate issue in #357

testkit/src/integration-testkit.ts
johanandren: Ah, indeed, we should only have `serviceName`.

sdk/src/action.ts
pvlugter: 👍🏼 Having it on the `context` could make more sense.

docs/src/modules/javascript/pages/access-control.adoc
```diff
@@ -0,0 +1,42 @@
+= Access Control Lists (ACLs)
+
+This section describes functionality in the JavaScript/TypeScript SDK related to ACLs (Access Control Lists), for information about how the ACLs are defined and control access see https://docs.kalix.io/services/using-acls.html[Using ACLs]
+
+== Default ACL in project templates
+
+If no ACLs is defined at all in a Kalix service Kalix will allow requests from both other services and the internet to all components of a Kalix service.
```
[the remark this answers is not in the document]
pvlugter: Did the small changes in the docs get lost after force push?

PR #356, opened 2022-06-09
Add SDK version docs update to release template

beritou:


ennru (approved):
Great to add that here!

pvlugter:
Just a note that we shouldn't really have that version in the main docs (also not needed for JVM SDKs). The latest version attribute is automatically written when publishing the docs:

https://github.com/lightbend/kalix-javascript-sdk/blob/9edd876821f128aa9f1950def45873896190a0e3/docs/Makefile#L31-L32

Would be better to use that one, rather than a manually managed version. May still be there because of the JS quickstarts being in the main docs rather than in the SDK docs.

Issue #99, opened 2021-07-07
Update to support M1 architecture


kikiya:
Currently the install fails when using on M1 architecture Macs but the current fix is pretty straight forward:

Based on issue https://github.com/protocolbuffers/protobuf/pull/8557 

We should be able to add these lines and might want to reference the above PR in comments. 

```
case "arm64":
          return makeDownloadFile("osx-x86_64"); 
```

The change should be added in this block -- 

https://github.com/lightbend/akkaserverless-javascript-sdk/blob/69f5578a000b990d02d685ffcf8514e89d985e48/sdk/bin/download-protoc.js#L17

johanandren:
One more step is some details about building the docker image using `docker buildx` with amd64 as platform.

johanandren:
Adding that osx-x86_64 then goes on to fail with 
```
[ 66%] Building C object CMakeFiles/cpu_features.dir/src/cpuinfo_arm.c.o
In file included from /Users/johan/Code/Lightbend/akka-serverless/akkaserverless-javascript-sdk/sdk/node_modules/cpu-features/deps/cpu_features/src/cpuinfo_arm.c:15:
/Users/johan/Code/Lightbend/akka-serverless/akkaserverless-javascript-sdk/sdk/node_modules/cpu-features/deps/cpu_features/include/cpuinfo_arm.h:118:2: error: "Including cpuinfo_arm.h from a non-arm target."
#error "Including cpuinfo_arm.h from a non-arm target."
 ^
```

in "cpu_features" https://github.com/google/cpu_features/pull/150

So looks like moving forward with this is blocked by that

johanandren:
Looks like this _may_ be fixed by the switch in what proto/grpc tooling we use in current main.

johanandren:
Almost, it is not possible to build the kalix-tools because it depends on a native image of the codegen which is not available for Macos arm64.

johanandren:
I got that to work locally by doing the following:
 * install a graalvm that works on m1 ("22.1.0.r17-grl" with sdkman)
 * point the `nativeImageGraalHome` in `kalix-codegen-js-cli` to the graal directory
 * sbt kalix-codegen-js-cli/nativeImage
 * copy the built binary somewhere
 * set/export env KALIX_NPMJS_CODEGEN_BINARY to the complete path of that somewhere

pvlugter:
Looks like CircleCI doesn't yet support M1, so we won't be able to automate this yet.

[...]

PR #358, opened 2022-06-16
bump: updates for apple silicon

pvlugter:
Refs #99.

I don't have an M1 mac to try this out on, but trying out some updates.

Bump protoc to a version which has an `osx-aarch_64` download. Not strictly needed if running the x86 version with rosetta works anyway.

Bump `nativeImageVersion` for building locally. Should automatically download an M1-compatible graalvm.

CircleCI doesn't support M1 runners yet, so we can't automatically publish the codegen binaries. Use the x86 download with rosetta instead for now.

@johanandren, can you check whether this works? We can also make a release if it's easier to test in a sample properly.


pvlugter:
I'll merge and release this, so it can be tested easily on a template.

PR #362, opened 2022-06-22
doc: standardize Quickstart code conventions

beritou:
Migrate Event Sourced Entity Quickstart module format from ES to
CommonsJS. This standardizes both the Value Entity Quickstart and
Event Sourced Entity Quickstart on the same module format.

Refs: https://github.com/lightbend/kalix/issues/6977

### DIFF
--- a/samples/js/js-eventsourced-shopping-cart/src/index.js
+++ b/samples/js/js-eventsourced-shopping-cart/src/index.js
@@ -14,15 +14,11 @@
  * limitations under the License.
  */
 // tag::imports[]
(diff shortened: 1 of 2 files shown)


pvlugter (commented):
Could be better to standardise on ES modules instead, as that's what the codegen is generating. And could be simpler if the quickstarts were based on the templates and codegen anyway?

Fails the tests.

beritou:
After reviewing this further, I agree, standardizing on ES modules would be better. Closing this PR as it is no longer needed.

PR #363, opened 2022-06-22
feat: support TypeScript codegen and templates

### LINKED ISSUE #163: TypeScript support for codegen and kickstart

pvlugter:
Resolves #163.

Add support for generating TypeScript sources and templating TypeScript projects.

Shares as much as possible. Since the codegen is so similar, just switches within the current JavaScript codegen, rather than duplicating everything. Similarly for templates. Updates the TypeScript samples that were using `kalix-scripts` already (and deleting the generated JavaScript files).


docs/src/modules/javascript/pages/kickstart.adoc
```diff
@@ -51,6 +56,30 @@ yarn create @kalix-io/kalix-entity@latest my-entity
...
+----
+--
+yarn:::
++
+--
+[source, command window, subs="attributes"]
+----
+yarn create @kalix-io/kalix-entity@latest my-entity --typescript
+----
+--
```
pvlugter, line 77: Nested dlists work, but the styling for tabs may not be setup for this.

npm-js/create-kalix-entity/template/base-ts/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+COPY --from=builder --chown=node /home/node/proto ./proto
+COPY --from=builder --chown=node /home/node/dist ./dist
+
+# Run the app as an unprivileged user for extra security.
+USER node
+
+# Run
+EXPOSE 8080
+# Call node directly to get SIGTERM for graceful shutdown
+CMD ["node", "dist/src/index.js"]
```
pvlugter, line 47: Easiest to test the Dockerfile once this is released, so will need to follow up with checking this.

npm-js/kalix-scripts/bin/kalix-scripts.js
```diff
@@ -76,14 +113,16 @@ const scriptHandlers = {
...
+        ...flag('--typescript', typescript),
+        ...option('--base-dir', baseDir),
+        ...option('--main-file', mainFile),
+        ...option('--descriptor-set-output-dir', descriptorSetOutputDir),
+        ...option('--descriptor-set-file', descriptorSetFile),
+        ...option('--proto-source-dir', protoSourceDir),
+        ...option('--source-dir', sourceDir),
+        ...option('--generated-source-dir', generatedSourceDir),
+        ...option('--test-source-dir', testSourceDir),
+        ...option('--integration-test-source-dir', integrationTestSourceDir),
```
pvlugter, line 125: Support all of the possible options in the config. I think it could also be good to support a separate config json file. And the `kalix-scripts build` could be used for the testkit example project now. Will follow up on these separately, and any other codegen changes.

johanandren (approved):
LGTM (with the follow ups followed up)

docs/src/modules/javascript/pages/kickstart.adoc
pvlugter: Updated to not use nested dlists in #370

npm-js/create-kalix-entity/template/base-ts/Dockerfile
pvlugter: Tried this out and docker build for TypeScript projects looks fine.

PR #368, opened 2022-06-27
build: update testkit publishing

pvlugter:
This should avoid the build failures on testkit publish by always building against a local snapshot of the SDK. For when the testkit has been updated and depends on corresponding changes in the SDK.

### DIFF
--- a/testkit/package.json
+++ b/testkit/package.json
@@ -31,13 +31,13 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "peerDependencies": {
-    "@kalix-io/kalix-javascript-sdk": "^1.0.0-M7"
+    "@kalix-io/kalix-javascript-sdk": "1.x"
   },
   "dependencies": {
     "testcontainers": "^8.10.1"
   },
   "devDependencies": {
-    "@kalix-io/kalix-javascript-sdk": "^1.0.0-M7",
+    "@kalix-io/kalix-javascript-sdk": "^1.0.0-M10",
     "@types/chai": "^4.3.1",
     "@types/mocha": "^9.1.1",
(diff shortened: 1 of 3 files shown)


testkit/package.json
```diff
@@ -31,13 +31,13 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "peerDependencies": {
-    "@kalix-io/kalix-javascript-sdk": "^1.0.0-M7"
+    "@kalix-io/kalix-javascript-sdk": "1.x"
```
pvlugter, line 34: The expectation is for the SDK and testkit to be on the same version. The peer dependency on the SDK is simplified to just be `1.x`, as this would need to be manually updated — although not necessarily accurate if the testkit depends on newly introduced features in a minor version of the SDK. We could otherwise drop the peer dependency altogether.

testkit/package.json
```diff
@@ -31,13 +31,13 @@
...
   },
   "dependencies": {
     "testcontainers": "^8.10.1"
   },
   "devDependencies": {
-    "@kalix-io/kalix-javascript-sdk": "^1.0.0-M7",
+    "@kalix-io/kalix-javascript-sdk": "^1.0.0-M10",
```
pvlugter, line 40: CI always installs a local snapshot for testing, and now for publishing. But given that we're using the package lock file, and `npm ci`, for other dependencies, this could be updated manually at times to a compatible version. Note that the semantic versioning probably doesn't support the `M{n}` format, but requires `m.{n}` or similar, with a dot-separated numeric part. But this should work better once we're past 1.0.0 final.

pvlugter:
Also created an issue on whether we should update the build tooling used: #369

Issue #235, opened 2022-01-07
[epic] New codegen annotation


octonato:
- [x] add new annotations in SDK project #230
- [ ] add more tests that verifies codegen is producing value code
- [ ] adapt documentation
- [ ] samples
- [ ] adapt TCK tests in the SDK
- [ ] adapt quickstarts

PR #373, opened 2022-06-28
fix: enable codegen with new annotation

pvlugter:
Refs #235. Allow the new codegen annotation to be used (just missing extension registration). There are no tests for this yet, but currently trying it out in updated quickstarts.


codegen/core/src/main/scala/io/kalix/codegen/DescriptorSet.scala
```diff
@@ -42,6 +42,7 @@ object DescriptorSet {
     Using[FileInputStream, Either[CannotOpen, Iterable[Either[ReadFailure, Descriptors.FileDescriptor]]]](
       new FileInputStream(file)) { fis =>
       val registry = ExtensionRegistry.newInstance()
+      registry.add(kalix.Annotations.codegen)
```
pvlugter, line 45: Was just missing this registration. Model building already added.

PR #375, opened 2022-06-28
chore: remove unnecessary dependencies from templates

pvlugter:
Not sure why `jsdoc` and `uglify-js` were listed as dependencies in the templates, but shouldn't be needed as explicit dependencies.


pvlugter:
These are also in samples, which can be looked at when those are next updated.

johanandren (approved):

PR #379, opened 2022-06-29
docs: add new quickstart for customer registry in javascript

pvlugter:
The current customer registry sample includes more than needed for the quickstart (has both value entity and event sourced entity versions) and is not set up for being a usable and runnable download. The quickstart docs for this sample (currently in the main docs) are outdated, use line numbers for code samples, and don't use the codegen.

Update things in line with the customer registry quickstarts for JVM SDKs.

Create a new version of the customer registry sample for the quickstart, based on the template and using codegen, with tests, and its own quickstart-friendly README. The quickstart bundle is tested in CI. This is a new customer registry sample, with the old one unchanged (still referenced by the main docs), and the old sample can be updated later (like in the JVM SDKs).

Add a new quickstart guide to the docs here, updated to use the basic template and codegen. The one in the main docs can be replaced once this is published.

Once reviewed, a TypeScript version can be added. The other quickstart (shopping cart) is being updated in https://github.com/lightbend/kalix-docs/pull/1402, which can also be moved here and its quickstart download updated.

### DIFF
--- a/samples/js/js-customer-registry-quickstart/.gitignore
+++ b/samples/js/js-customer-registry-quickstart/.gitignore
@@ -1,0 +1,5 @@
+/lib/generated
+/node_modules
+/user-function.desc
+# test quickstarts without lock file
+/package-lock.json
--- a/samples/js/js-customer-registry-quickstart/proto/customer_api.proto
+++ b/samples/js/js-customer-registry-quickstart/proto/customer_api.proto
@@ -1,0 +1,74 @@
+// Copyright 2021 Lightbend Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// tag::declarations[]
+syntax = "proto3";
+
+package customer.api;
+
+import "google/protobuf/empty.proto";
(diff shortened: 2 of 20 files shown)


samples/js/js-customer-registry-quickstart/.gitignore
```diff
@@ -0,0 +1,5 @@
+/lib/generated
+/node_modules
+/user-function.desc
+# test quickstarts without lock file
+/package-lock.json
```
pvlugter, line 5: Not committing the package lock file for this sample, so that testing is similar to user download.

samples/js/js-customer-registry-quickstart/proto/customer_api.proto
```diff
@@ -0,0 +1,74 @@
...
+  };
+  // end::service[]
+  option (kalix.service).acl.allow = { principal: ALL };
+  // tag::service[]
+
+  rpc Create(Customer) returns (google.protobuf.Empty) {}
+  rpc GetCustomer(GetCustomerRequest) returns (Customer) {}
+  // end::service[]
+  rpc ChangeName(ChangeNameRequest) returns (google.protobuf.Empty) {}
+  rpc ChangeAddress(ChangeAddressRequest) returns (google.protobuf.Empty) {}
```
pvlugter, line 71: This has the name and address changing methods, like for the Java/Scala versions, but these are not included in the quickstart guide in the docs.

johanandren (approved):
LGTM!

samples/js/js-customer-registry-quickstart/proto/customer_api.proto
```diff
@@ -0,0 +1,74 @@
...
+
+// tag::service[]
+service CustomerService {
+  option (kalix.codegen) = {
+    value_entity: {
+      name: "customer.domain.Customer"
+      entity_type: "customers"
+      state: "customer.domain.CustomerState"
+    }
+  };
```
pvlugter, line 62: Also using the new codegen annotation now.

PR #381, opened 2022-07-01
docs: add new shopping cart quickstarts

pvlugter:
Refs https://github.com/lightbend/kalix/issues/6959

Add new shopping cart quickstart samples and docs for both JavaScript and TypeScript. This is a blend of the existing sample and quickstart guide in the main docs, and the quickstarts for JVM SDKs. Again, added as new samples with the old samples left as is for now.

The TypeScript version is more defensive than needed, as the types it's currently working with are `| null | undefined` optional, while the actual deserialised types have defaults set. This is issue #326, and something to come back to here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,6 +273,21 @@ jobs:
             npm run build
             npm test
       - run:
+          name: "Test quickstart: TS Shopping Cart"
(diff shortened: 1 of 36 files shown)


johanandren (approved):
LGTM!

PR #388, opened 2022-07-05
bump: kalix-proxy 1.0.12

patriknw:


pvlugter:
Hmm, CI doesn't seem to like the new proxy image...

pvlugter:
Fixed the external link validation in #390.

The issues with the docker images look odd. TCK tests failing with:

```
No java installations was detected.
Please go to http://www.java.com/getjava/ and download
```

pvlugter:
So looks like the temurin on ubuntu jammy that we're using as the new base image for the proxy is broken on older docker versions. This is https://github.com/adoptium/containers/issues/215 and other issues. Updated to latest circleci machin image to get newer version of docker and everything else in #391. Manually checking it on a CI box and that looks fine. Probably want to bump the minimum docker version in the docs as well — not sure exactly what the min version is now, but 20.10.14 works in CI.

pvlugter (approved):
Working now, on newer docker version

PR #391, opened 2022-07-07
ci: update machine image

pvlugter:
Latest proxy docker images (on temurin 17) are not usable on older docker. Update the machine image for CircleCI to get newer versions of everything. I'll manually check this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,7 +179,7 @@ jobs:
 
   tests:
     machine:
-      image: ubuntu-2004:202104-01
+      image: ubuntu-2204:2022.04.2
     description: "tests"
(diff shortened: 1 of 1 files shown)


pvlugter:
Docker version used now (20.10.14) looks fine for the temurin 17 on jammy images.

pvlugter:
That also bumped the provided node and npm versions, which then breaks some installs. Also add `nvm install` to the builds and tests that didn't have this already. Could make sense to a bump to the next node LTS version as well (currently on the old maintenance LTS version 14, current LTS is 16).

pvlugter:
As CI is passing now, with old proxy images, will merge now to try #388.

PR #393, opened 2022-07-12
fix: add severity to user function error reporting

pvlugter:
Refs https://github.com/lightbend/kalix-proxy/issues/1566. Builds on #392.

### DIFF
--- a/sdk/src/kalix.ts
+++ b/sdk/src/kalix.ts
@@ -560,13 +560,7 @@ export class Kalix {
         }
       },
       ReportError(call, callback) {
-        const msg = self.reportErrorLogic(
-          call.request.code,
-          call.request.message,
-          call.request.detail,
-          call.request.sourceLocations,
-        );
-        console.error(msg);
+        self.reportError(call.request);
         callback(null, {});
       },
       ProxyTerminated(_call, callback) {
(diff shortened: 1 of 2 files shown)


sdk/src/kalix.ts
```diff
@@ -605,22 +599,54 @@ export class Kalix {
...
+      error.code,
+      error.message,
+      error.detail,
+      error.sourceLocations,
+      error.severity,
+    );
+    switch (error.severity) {
+      case discovery.Severity.INFO:
+      case discovery.Severity.WARNING:
+        console.log(msg);
```
pvlugter, line 614: There is a `console.warn` but it's an alias for `console.error`, which prints to stderr. This puts warnings on stdout with `console.log` instead, as GCP logging will mark stderr as ERRORs. But maybe having it use `console.warn` and print to stderr is better anyway?

sdk/src/kalix.ts
```diff
@@ -278,6 +278,14 @@ export interface ServiceMap {
   [key: string]: protobuf.Service;
 }
 
+/** @internal */
+enum UserFunctionErrorSeverity {
+  UNSPECIFIED = 0,
+  ERROR = 1,
+  WARNING = 2,
+  INFO = 3,
+}
```
pvlugter, line 287: Forgot that the protocol types can only be used in type position. Using an enum compiles in a reference to it, but these types are not packaged in the dist. So mirror the severity enum here for switches. Seems a bit awkward, so maybe the protocol typing and whether that gets included in the package, should be revisited.

sdk/src/kalix.ts
```diff
@@ -605,22 +607,54 @@ export class Kalix {
...
+      error.code,
+      error.message,
+      error.detail,
+      error.sourceLocations,
+      error.severity,
+    );
+    switch (error.severity as number) {
+      case UserFunctionErrorSeverity.INFO:
+      case UserFunctionErrorSeverity.WARNING:
+        console.log(msg);
```
pvlugter, line 622: There is a `console.warn` but it's an alias for `console.error`, which prints to stderr. This puts warnings on stdout with `console.log` instead, as GCP logging will mark stderr as ERRORs. But maybe having it use `console.warn` and print to stderr is better anyway?

pvlugter:
Rebased and ready for review now.

sdk/src/kalix.ts
johanandren: If we don't have any more fine grained output then I think console.warn/stderr makes sense.

sdk/src/kalix.ts
```diff
@@ -605,22 +607,54 @@ export class Kalix {
...
+    switch (severity as number) {
+      case UserFunctionErrorSeverity.INFO:
+        messageType = 'Message';
+        break;
+      case UserFunctionErrorSeverity.WARNING:
+        messageType = 'Warning';
+        break;
+    }
+
+    let msg = `${messageType} reported from Kalix system: ${code} ${message}`;
```
johanandren, line 648: BTW, shouldn't this be JSON anyway?

johanandren (approved):

sdk/src/kalix.ts
pvlugter: Right, no structured JSON logging anywhere at the moment. Older issue is https://github.com/lightbend/kalix-proxy/issues/618. Could make sense to have our own small logging API, that users can also make use of, from the context or something.

sdk/src/kalix.ts
pvlugter: Ok, will update to use console.{info,warn,error}. And we can then switch to a proper logging layer later, with structured JSON output. Follow up issue will be https://github.com/lightbend/kalix-proxy/issues/618.

Issue #394, opened 2022-07-12
Partial download of codegen


johanandren:
Created shopping cart quickstart and tried to build it

On Linux x86_64:

```
/srv/homes/johan/code/lightbend/shopping-cart/node_modules/@kalix-io/kalix-scripts/bin/kalix-codegen-js.bin --typescript --proto-source-dir ./proto --source-dir ./src --generated-source-dir ./lib/generated --test-source-dir ./test --integration-test-source-dir ./integration-test
Segmentation fault
```

On darvin arm64: 
```
/Users/johan/Code/Lightbend/Kalix/shopping-cart-quickstart/node_modules/@kalix-io/kalix-scripts/bin/kalix-codegen-js.bin --typescript --proto-source-dir ./proto --source-dir ./src --generated-source-dir ./lib/generated --test-source-dir ./test --integration-test-source-dir ./integration-test
fish: Job 1, '/Users/johan/Code/Lightbend/Kal…' terminated by signal SIGKILL (Forced quit)
```

johanandren:
Didn't figure out a way to detect this, body is a stream so would have to be comparing `response.headers.get('content-length')` with the bytes piped through when done writing or something.

pvlugter:
That's unpleasant. Seems to fail the downloads quite often.

Have also had a play around with detecting. Couldn't get any errors from the stream until trying it on Node 16, where it will signal this error event on the response body:

```
Error: aborted
    at connResetException (node:internal/errors:692:14)
    at TLSSocket.socketCloseListener (node:_http_client:414:19)
    at TLSSocket.emit (node:events:539:35)
    at node:net:709:12
    at TCP.done (node:_tls_wrap:582:7) {
  code: 'ECONNRESET'
}
```

Following the changes there, it seems that it should have an `aborted` event on Node 14. But don't see that emitted for the body stream with node-fetch. Trying out axios in place of node-fetch and can get the aborted signal (seems it's a different type of stream as well, `IncomingMessage` instead of `PassThrough`).

Can also use `stream.pipeline` to have error handling attached automatically, and then it has this error:

```
Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close
    at new NodeError (internal/errors.js:322:7)
    at IncomingMessage.onclose (internal/streams/end-of-stream.js:117:38)
    at IncomingMessage.emit (events.js:400:28)
    at TLSSocket.socketCloseListener (_http_client.js:432:11)
[...]

PR #395, opened 2022-07-13
fix: handle disconnected streams on codegen download

pvlugter:
Refs #394

Switch over to axios, which gets the disconnected signals. And use `stream.pipeline` which adds error handling to the stream automatically (otherwise it's adding event handlers for `error` and `aborted` on the response data, and `close` on the writer for resolving successfully).

This just logs the error currently, on disconnected streams. Looks like this

```
Fetching kalix-codegen-js from https://repo.lightbend.com/raw/kalix/versions/1.0.0/kalix-codegen-js-x86_64-apple-darwin
Saving to .../kalix-scripts/bin/kalix-codegen-js.bin
Failed to download Kalix codegen binary: Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close
    at new NodeError (internal/errors.js:322:7)
    at IncomingMessage.onclose (internal/streams/end-of-stream.js:117:38)
    at IncomingMessage.emit (events.js:400:28)
    at TLSSocket.socketCloseListener (_http_client.js:432:11)
    at TLSSocket.emit (events.js:412:35)
    at net.js:686:12
    at TCP.done (_tls_wrap.js:564:7) {
  code: 'ERR_STREAM_PREMATURE_CLOSE'
}
```

Would make sense to wrap in some retries.

### DIFF
--- a/npm-js/kalix-scripts/bin/download-codegen.js
+++ b/npm-js/kalix-scripts/bin/download-codegen.js
@@ -1,6 +1,8 @@
-const fetch = require('node-fetch');
+const axios = require('axios').default;
 const fs = require('fs');
 const path = require('path');
+const stream = require('stream');
+const util = require('util');
 const packageConfig = require('../package.json');
 
 /**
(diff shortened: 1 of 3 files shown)


pvlugter:
The underlying issue with the proxy load balancing seems to be resolved, but it may still be useful to have the error handling here. Not sure we should worry about retries, but it does at least detect and show the error, and remove the partial download so it can't be used in a broken way.

pvlugter:
Also add a process exit on errors, so that scripts don't continue past this point and it's clear it failed.

ennru (approved):
LGTM.

PR #396, opened 2022-07-15
chore: avoiding calling through the mesh to same services components

franciscolopezsancho:
References https://github.com/lightbend/kalix-proxy/issues/1469 and depends on https://github.com/lightbend/kalix-proxy/pull/1599

### DIFF
--- a/sdk/src/kalix.ts
+++ b/sdk/src/kalix.ts
@@ -717,7 +717,7 @@ export class Kalix {
         proxyHostname:
           proxyInfo.internalProxyHostname ?? proxyInfo.proxyHostname,
         proxyPort:
-          proxyInfo.proxyPort ??
+          this.proxyPort ??
           (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80),
         identificationInfo: proxyInfo.identificationInfo || undefined,
       };
--- a/sdk/src/kalix.ts
+++ b/sdk/src/kalix.ts
@@ -698,10 +698,11 @@ export class Kalix {
       this.proxyHasTerminated = false;
 
       console.log(
-        'Received discovery call from [%s %s] at [%s] supporting Kalix protocol %s.%s',
+        'Received discovery call from [%s %s] at [%s]:[%s] supporting Kalix protocol %s.%s',
         proxyInfo.proxyName,
         proxyInfo.proxyVersion,
-        proxyInfo.proxyHostname,
+        proxyInfo.internalProxyHostname,
+        proxyInfo.proxyPort,
         proxyInfo.protocolMajorVersion,
         proxyInfo.protocolMinorVersion,
       );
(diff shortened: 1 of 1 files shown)


sdk/src/grpc-util.ts
```diff
@@ -62,8 +62,7 @@ export interface GrpcClientLookup {
  * @private
  */
 enum ServiceType {
-  SELF,
-  KALIX_SERVICE,
```
johanandren, line 66: This change is not good, the possible clients a user wants to look up are from the service itself (`SELF`) other deployed Kalix services in the same project (`KALIX_SERVICE`) and arbitrary gRCP services (`EXTERNAL`).

For the first two we need to send different set of auth headers. For external services we send no auth headers at all.

johanandren (dismissed):

sdk/src/grpc-util.ts
franciscolopezsancho: Interesting. Thanks for the heads up. I expected something to break and alert me if this was really necessary. You saved me some precious time.

franciscolopezsancho:
I think I understand. When it tries to reach an gRPC from the same Kalix service uses some headers. I think it advertises that is `self`. But when it reaches another gRPC from another Kalix service it signals that is some other service. Or maybe the lack of `self` is sufficient. And finally when reaching gRPC not of a Kalix service doesn't say whether if it is self or not self.

johanandren:
You can see the different pairs of headers for self vs other Kalix services coming from the proxy and passed on to each component here:

https://github.com/lightbend/kalix-javascript-sdk/blob/main/sdk/src/kalix.ts#L684

Then used in GrpcUtils here:
https://github.com/lightbend/kalix-javascript-sdk/blob/main/sdk/src/grpc-util.ts#L155
and here:
https://github.com/lightbend/kalix-javascript-sdk/blob/main/sdk/src/grpc-util.ts#L159

sdk/src/kalix.ts
```diff
@@ -678,9 +679,7 @@ export class Kalix {
 
       const preStartSettings: PreStartSettings = {
         proxyHostname: proxyInfo.proxyHostname,
-        proxyPort:
-          this.proxyPort ??
-          (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80),
+        proxyPort: proxyInfo.proxyPort,
```
franciscolopezsancho, line 682: This might not be even needed. But I'm reluctant to use 9000 here as a hardcoded value that comes from Kalix proxy

sdk/src/grpc-util.ts
franciscolopezsancho: I put it back.

sdk/src/kalix.ts
octonato: The code is correct, we should use whatever port the proxy send to us. Not 9000.

octonato (approved):
LGTM

octonato:
See https://github.com/lightbend/kalix-jvm-sdk/pull/1055/files#r925790190

octonato:
commits squashed and rebased with #400. 

Waiting for #400

octonato:
Depends on #400 (new proxy version)

efgpinto:
#400 is merged now

octonato:
This is now rebased with main (use new proxy version) and is ready for merge. 

Johan had a change request, but Cisco has already implemented it.

pvlugter:
CI is failing on the testkit's own integration tests, and looks like a relevant failure:

```
1) The Kalix IntegrationTestkit
       should handle self-delegating actions (with acl):
     Error: 2 UNKNOWN: Cross component call failed
```

franciscolopezsancho:
I'll have a look

pvlugter:
@franciscolopezsancho, with the change from `proxyInfo.proxyPort` back to `this.proxyPort`, does that mean we're not using the proxy port passed from the proxy any longer — breaking the feature, especially in actual deployments?

franciscolopezsancho:
You are right I fixed it in https://github.com/lightbend/kalix-javascript-sdk/pull/410. Please have a look.

PR #403, opened 2022-08-08
fix: modify pbjs and pbts path

### LINKED ISSUE #402: wrong path for pbjs and pbts in our sample app
Customer reported that they got some error when following the [quick start guide](https://docs.kalix.io/javascript/quickstart/cr-value-entity-javascript.html)

Here are steps to reproduce the problem....

```
# download sample
kalix quickstart download customer-registry-javascript

# build the sample
cd customer-registry
npm install
npm run build
```

We will see the following errors
```
/bin/sh: /xxx/customer-registry/node_modules/.bin/pbjs: No such file or directory
```
There are two possible solutions:

### solution 1:
Add `protobufjs@6.11.3` as devDependencies in  in https://github.com/lightbend/kalix-javascript-sdk/blob/main/samples/js/js-customer-registry-quickstart/package.json 

### solution 2:
Modify the `pbjs` path in kalix-script.js to `./node_modules/@grpc/proto-loader/node_modules/.bin/pbjs`, and the similar to `pbts`
https://github.com/lightbend/kalix-javascript-sdk/blob/main/npm-js/kalix-scripts/bin/kalix-scripts.js#L103-L115

One thing I feel  confused is that, in https://github.com/lightbend/kalix-javascript-sdk/blob/main/samples/js/js-customer-registry-quickstart/package.json  It uses `kalix-script` in build. However, in https://github.com/lightbend/kalix-javascript-sdk/blob/main/samples/js/js-customer-registry/package.json , it uses `compile-descriptor` for build instead of `kalix-script`, why are they different???

@jroper and @octonato , I know you are working on this area recently , especially the issue https://github.com/lightbend/kalix-javascript-sdk/issues/398   Any comment on this issue?

katsutoxin:
Fix #402

### DIFF
--- a/npm-js/kalix-scripts/bin/kalix-scripts.js
+++ b/npm-js/kalix-scripts/bin/kalix-scripts.js
@@ -102,14 +102,14 @@ const scriptHandlers = {
 
     runOrFail(
       'Building static Javascript definitions from proto',
-      path.resolve('node_modules', '.bin', 'pbjs'),
+      path.resolve('node_modules', '@grpc', 'proto-loader', 'node_modules','.bin', 'pbjs'),
       [...protoFiles, '-t', 'static-module', '-o', protoJs],
       { shell: true },
     );
 
     runOrFail(
       'Building Typescript definitions from static JS',
-      path.resolve('node_modules', '.bin', 'pbts'),
+      path.resolve('node_modules', '@grpc', 'proto-loader', 'node_modules','.bin', 'pbts'),
       [protoJs, '-o', protoTs],
       { shell: true },
     );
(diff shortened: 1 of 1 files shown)


katsutoxin:
Looks like if I change the path, the integration tests will failed 

https://app.circleci.com/pipelines/github/lightbend/kalix-javascript-sdk/1273/workflows/3a23a7d2-84b4-4bf7-bf09-f51400b301eb/jobs/8193

katsutoxin:
cc @pvlugter ,  I feel that we might need to change the path for pbts.....   Please let me know if we need this fix.

pvlugter:
Not needed. Should work fine now that the gprc-js and proto-loader versions are pinned. Have released 1.0.2.

PR #406, opened 2022-08-11
ci: update machine images

pvlugter:
* downgrade ubuntu images from 22.04 to latest 20.04
* upgrade macos image to 12.5.1

Fixes following the 1.0.1 release (#402).

Codegen binary for linux was built against a newer glibc (building with ubuntu 22.04) than in the docker image (debian 10). The Debian 11 (node bullseye) images are still on older glibc as well. To keep things simple, downgrade the machine images to ubuntu 20.04 again, but should still be with a new enough docker version (the reason they were bumped originally, #391).

The 1.0.1 release failed to build the macOS codegen binary, as 12.5.0 image was not available. Bump to available 12.5.1.

I'll manually confirm a couple of things in the test build.

Would be useful to have some tests that build these, and check across platforms, but for another time...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,7 +179,7 @@ jobs:
 
   tests:
     machine:
-      image: ubuntu-2204:2022.04.2
+      image: ubuntu-2004:2022.07.1
     description: "tests"
     steps:
       - checkout-and-merge-to-main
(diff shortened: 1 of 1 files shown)


pvlugter:
This seems good. It's for the publish jobs, so currently need to run a release to check. I'll merge and release again to confirm.

PR #410, opened 2022-08-17
feat: Avoiding the mesh

franciscolopezsancho:
References https://github.com/lightbend/kalix-proxy/issues/1469 and depends on https://github.com/lightbend/kalix-proxy/pull/1599

Continuation of https://github.com/lightbend/kalix-javascript-sdk/pull/396

### DIFF
--- a/sdk/src/kalix.ts
+++ b/sdk/src/kalix.ts
@@ -719,7 +719,7 @@ export class Kalix {
         proxyPort:
           this.proxyPort ??
           (proxyInfo.proxyPort ||
-          (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80)),
+            (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80)),
         identificationInfo: proxyInfo.identificationInfo || undefined,
       };
 
--- a/sdk/src/kalix.ts
+++ b/sdk/src/kalix.ts
@@ -714,7 +714,7 @@ export class Kalix {
       );
 
       const preStartSettings: PreStartSettings = {
-        proxyHostname: proxyInfo.proxyHostname,
+        proxyHostname: proxyInfo.internalProxyHostname,
         proxyPort: proxyInfo.proxyPort,
         identificationInfo: proxyInfo.identificationInfo || undefined,
       };
--- a/sdk/src/kalix.ts
+++ b/sdk/src/kalix.ts
@@ -718,8 +718,8 @@ export class Kalix {
           proxyInfo.internalProxyHostname ?? proxyInfo.proxyHostname,
         proxyPort:
           this.proxyPort ??
-          proxyInfo.proxyPort ||
-          (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80),
+          (proxyInfo.proxyPort ||
+          (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80)),
         identificationInfo: proxyInfo.identificationInfo || undefined,
       };
 
--- a/sdk/src/kalix.ts
+++ b/sdk/src/kalix.ts
@@ -662,10 +662,11 @@ export class Kalix {
       this.proxyHasTerminated = false;
 
       console.log(
-        'Received discovery call from [%s %s] at [%s] supporting Kalix protocol %s.%s',
+        'Received discovery call from [%s %s] at [%s]:[%s] supporting Kalix protocol %s.%s',
         proxyInfo.proxyName,
         proxyInfo.proxyVersion,
         proxyInfo.proxyHostname,
+        proxyInfo.proxyPort,
         proxyInfo.protocolMajorVersion,
         proxyInfo.protocolMinorVersion,
       );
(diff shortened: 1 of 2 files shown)


sdk/src/kalix.ts
```diff
@@ -717,6 +717,7 @@ export class Kalix {
         proxyHostname:
           proxyInfo.internalProxyHostname ?? proxyInfo.proxyHostname,
         proxyPort:
+          proxyInfo.proxyPort ??
           this.proxyPort ??
           (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80),
```
pvlugter, line 722: Probably should take into account `this.proxyPort` first.

Also, I'd expect `proxyInfo.proxyPort` will be 0 when it's not set (default value) if running an older version of the proxy. The `??` operator checks for null or undefined.

So I'd be inclined towards:

```suggestion
          this.proxyPort ??
          proxyInfo.proxyPort ||
          (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80),
```

pvlugter: Needs to be tested with the proxy, including before the internal hostname and port were added to the proxy.

Would be good if there was a test included for this...

franciscolopezsancho: I thought the tests we have already were testing the old implementation. Meaning no internal port nor port were in the `proxyInfo`.  Let me review that and in either case, add the test it is missing.

franciscolopezsancho: After reviewing the code I think that testing the current SDK against the old proxy is already implemented as `proxyPort`  or `proxyInternalHost` aren't set in the testkit. As [discoveryLogic](https://github.com/lightbend/kalix-javascript-sdk/blob/66126484b01ce7d6afa863e3f7fb8046d7bd7c2b/sdk/src/kalix.ts#L673) is never used in the testkit, neither [createDiscoveryHandlers](https://github.com/lightbend/kalix-javascript-sdk/blob/66126484b01ce7d6afa863e3f7fb8046d7bd7c2b/sdk/src/kalix.ts#L558). I didn't find any other meaningful test a could add.  I'll try to connect early tomorrow with you to make sure I understood properly what you @pvlugter  mean.

franciscolopezsancho: > After reviewing the code I think that testing the current SDK against the old proxy is already implemented as `proxyPort` or `proxyInternalHost` aren't set in the testkit. As [discoveryLogic](https://github.com/lightbend/kalix-javascript-sdk/blob/66126484b01ce7d6afa863e3f7fb8046d7bd7c2b/sdk/src/kalix.ts#L673) is never used in the testkit, neither [createDiscoveryHandlers](https://github.com/lightbend/kalix-javascript-sdk/blob/66126484b01ce7d6afa863e3f7fb8046d7bd7c2b/sdk/src/kalix.ts#L558). I didn't find any other meaningful test a could add. I'll try to connect early tomorrow with you to make sure I understood properly what you @pvlugter mean.

After some more thought. Do you think is valuable to allow the testkit an option to NOT `setProxyPort` and test that? Or allow the testkit to start the proxy in another port than 9000 and then pass the `proxyPort` as part of the `ProxyInfo`? I'm not sure yet they add value but that may be the case.

franciscolopezsancho:
- [ ]  Test old proxy (1.14) with this new JS-SDK - OK. 
- [ ]  Test new proxy (1.15) without old JS-SDK (without this PR) - OK
   _This second test was probably unnecessary but just in case._

franciscolopezsancho:
In Kubernetes
- [ ] Test old proxy (1.14) with this new JS-SDK - OK.

sdk/src/kalix.ts
```diff
@@ -718,7 +718,8 @@ export class Kalix {
           proxyInfo.internalProxyHostname ?? proxyInfo.proxyHostname,
         proxyPort:
           this.proxyPort ??
-          (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80),
+          (proxyInfo.proxyPort ||
+            (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80)),
```
patriknw, line 722: what does this mean? if the `proxyInfo.proxyPort` is defined it will use port 9000? shouldn't it use the value from the `proxyInfo.proxyPort`?

johanandren: Some duplication here, `variable ?? elsevalue` will use `variable` if truthy and `elsevalue` if `variable` is falsy (undefined, null, empty string etc), so the `proxyPort || elseExpression` just duplicates that. 

```this.proxyPort ?? (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80)```

Should be enough.

Comment (johanandren) [reply sdk/src/kalix.ts:722]: Ah, sorry, `??` is "nullish coalescing" so not quite the regular JS falsyness monkey business but an additional concept returning the value if it is for example `""` or `0` but never if `undefined` or `null`. Given that I think it should rather be only the right hand part to not be unnecessary hard to read: 

```
  (proxyInfo.proxyPort || (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80))
```

franciscolopezsancho: before there was `this.proxyPort ||
            (proxyInfo.proxyHostname == 'localhost' ? 9000 : 80)`

johanandren: Ah, I missed `this` in there, so that's when it comes from the integration testkit I guess. 👍 NVM my noise then.

franciscolopezsancho: :) that's not noise. Thanks for the input, Johan. Yes, that's used by the integration test kit.

franciscolopezsancho: what is weird is that I would swear I added a comment to your comment pointing exactly that. I may have missed hitting `comment` I guess.

johanandren (approved):

PR #411, opened 2022-08-17
fix: changing 'file' to 'codegen'

franciscolopezsancho:
References (https://github.com/lightbend/kalix-docs/issues/1474)


pvlugter:
The callout comment will now be out of sync with the actual code example:

https://docs.kalix.io/javascript/value-entity.html#_defining_the_proto_files

We need all of the samples and code snippets to be updated as well. And the new codegen annotations are only on the service, not in the domain files.

franciscolopezsancho:
I'll update them

PR #417, opened 2022-08-30
chore: updating samples from file to codegen

franciscolopezsancho:
References https://github.com/lightbend/kalix-docs/issues/1474 and https://github.com/lightbend/kalix-javascript-sdk/pull/411

### DIFF
--- a/samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_domain.proto
+++ b/samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_domain.proto
@@ -23,7 +23,7 @@ import "kalix/annotations.proto";
 // end::declarations[]
 // tag::proto_state[]
 // Describes how this domain relates to an event sourced entity
-option (kalix.file).event_sourced_entity = {
+option (kalix.codegen).event_sourced_entity = {
   name: "ShoppingCart"
   entity_type: "eventsourced-shopping-cart"
   state: "Cart"
--- a/samples/js/js-replicated-entity-shopping-cart/proto/shoppingcart_domain.proto
+++ b/samples/js/js-replicated-entity-shopping-cart/proto/shoppingcart_domain.proto
@@ -21,7 +21,7 @@ package com.example.shoppingcart.domain; // <1>
 
 // end::domain[]
 // TODO: codegen for Replicated Entities
-// option (kalix.file).replicated_entity = {
+// option (kalix.codegen).replicated_entity = {
 //   name: "ShoppingCart"
 //   entity_type: "shopping-cart"
 //   replicated_counter_map: {
--- a/samples/js/js-valueentity-shopping-cart/proto/shoppingcart_domain.proto
+++ b/samples/js/js-valueentity-shopping-cart/proto/shoppingcart_domain.proto
@@ -26,7 +26,7 @@ import "kalix/annotations.proto";
 
 // tag::proto_state[]
 // Describes how this domain relates to a value entity
-option (kalix.file).value_entity = {
+option (kalix.codegen).value_entity = {
   name: "ShoppingCart"
   entity_type: "shopping-cart"
   state: "Cart"
--- a/samples/js/valueentity-counter/proto/counter_domain.proto
+++ b/samples/js/valueentity-counter/proto/counter_domain.proto
@@ -4,7 +4,7 @@ package com.example.domain; // <1>
 
 import "kalix/annotations.proto"; // <2>
 
-option (kalix.file).value_entity = { // <3>
+option (kalix.codegen).value_entity = { // <3>
   name: "Counter" // <4>
   entity_type: "counter" // <5>
   state: "CounterState" // <6>
--- a/samples/ts/ts-eventsourced-shopping-cart/proto/shoppingcart_domain.proto
+++ b/samples/ts/ts-eventsourced-shopping-cart/proto/shoppingcart_domain.proto
@@ -22,7 +22,7 @@ package com.example.shoppingcart.domain;
 import "kalix/annotations.proto";
 // tag::proto_state[]
 // Describes how this domain relates to an event sourced entity
-option (kalix.file).event_sourced_entity = {
+option (kalix.codegen).event_sourced_entity = {
   name: "ShoppingCart"
   entity_type: "eventsourced-shopping-cart"
   state: "Cart"
--- a/samples/ts/ts-replicated-entity-shopping-cart/proto/shoppingcart_domain.proto
+++ b/samples/ts/ts-replicated-entity-shopping-cart/proto/shoppingcart_domain.proto
@@ -21,7 +21,7 @@ package com.example.shoppingcart.domain; // <1>
 
 // end::domain[]
 // TODO: codegen for Replicated Entities
-// option (kalix.file).replicated_entity = {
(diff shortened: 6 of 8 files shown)


samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_domain.proto
```diff
@@ -23,7 +23,7 @@ import "kalix/annotations.proto";
 // end::declarations[]
 // tag::proto_state[]
 // Describes how this domain relates to an event sourced entity
-option (kalix.file).event_sourced_entity = {
+option (kalix.codegen).event_sourced_entity = {
```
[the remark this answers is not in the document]
franciscolopezsancho: I'll need to change the docs accordingly because this has some tagging.

franciscolopezsancho:
@johanandren from https://docs.kalix.io/javascript/proto.html 
```
We recommend that you define your service API and events and data associated with components separately. This allows business logic to evolve independently of the public interface. This page walks you through elements in an example shoppingcart_api.proto file and the associated shoppingcart_domain.proto file.
```

On the other hand in the JVM samples we have the entities defined in the `.api`.  Just to double check we don't recommend the above anymore. Would you agree?

pvlugter:
> from https://docs.kalix.io/javascript/proto.html
> 
> ```
> We recommend that you define your service API and events and data associated with components separately. This allows business logic to evolve independently of the public interface. This page walks you through elements in an example shoppingcart_api.proto file and the associated shoppingcart_domain.proto file.
> ```
> 
> On the other hand in the JVM samples we have the entities defined in the `.api`. Just to double check we don't recommend the above anymore. Would you agree?

We should still have the recommended separation for the protobuf messages, with a separate domain proto file. It's only the kalix annotations that are no longer split over the two files.

Would make sense to align with the JVM docs, which still have the above recommendation as well.

https://docs.kalix.io/java/writing-grpc-descriptors-protobuf.html

franciscolopezsancho:
I missed that in the JVM docs. Thanks @pvlugter

franciscolopezsancho:
I made all the changes but the error we are getting here - mongo-related - in the CI I think has nothing to do with my changes. 
I downloaded a fresh repo from main and it raises the same error when `npm run build` the `ts-value-entity` created with `npm-js/create-kalix-entity/create-kalix-entity.js`. 

So far I've been unable to fix it.

johanandren (approved):

johanandren:
Hmm, yeah looks unrelated, what's up with something Mongo being a dependency at all? 

> node_modules/@types/mongodb/index.d.ts:1198:45 - error TS2344: Type 'TSchema' does not satisfy the constraint '{ [key: string]: any; }'.

franciscolopezsancho:
Good point. Maybe removing it would do 👍. I'll have a look to see if doesn't break something else.

franciscolopezsancho:
TODO: try adding
```
  compilerOptions: {
    "skipLibCheck": true,
    //...
  }
```
to tsconfig

npm-js/create-kalix-entity/template/base-ts/tsconfig.json
```diff
@@ -3,8 +3,9 @@
     "outDir": "./dist",
     "strict": true,
     "strictFunctionTypes": true,
+    "skipLibCheck": true, //TODO consider https://classic.yarnpkg.com/en/docs/selective-version-resolutions/
```
johanandren, line 6: I don't quite understand the consequences of this, @pvlugter do you?

Problem was that it fails on some type validation in a mongo library which we for some unclear reason depend on

pvlugter: Reverted this change to see if the tests are ok now. Tests look fine on main as well.

pvlugter: Tests passing fine without this now.

samples/js/js-eventsourced-shopping-cart/proto/shoppingcart_api.proto
```diff
@@ -53,6 +53,18 @@ message GetShoppingCart {
...
+    entity_type: "eventsourced-shopping-cart"
+    state: "Cart"
+    events: ".domain.ItemAdded"
+    events: ".domain.ItemRemoved"
+};
+  // end::proto_state[]
+// tag::service[]
   option (kalix.service) = {
     type: SERVICE_TYPE_ENTITY
     component: ".domain.ShoppingCart"
```
pvlugter, line 70: The `kalix.service` annotations should be removed as well. The change is from having a `kalix.service` annotation connected to a `kalix.file` annotation with the component configuration, to just having a `kalix.codegen` annotation on the service with all the configuration.

franciscolopezsancho: Wow, that was weird. I remember reproducing in local but anyways. We are good now 👍  Thanks guys
Ok. I will remove `kalix.service`

npm-js/create-kalix-entity/template/event-sourced-entity/proto/counter_api.proto
```diff
@@ -30,9 +30,13 @@ message CurrentCounter {
...
-  option (kalix.service) = {
-    type : SERVICE_TYPE_ENTITY
-    component : ".domain.Counter"
+  option (kalix.codegen).event_sourced_entity = {
+    name: "Counter"
+    entity_type: "counter"
+    state: ".domain.CounterState"
+    events: ".domain.ValueIncreased"
+    events: ".domain.ValueDecreased"
+    events: ".domain.ValueReset"
```
pvlugter, line 39: We may want to use fully qualified names here, as the relative ones have tripped people before when they've also used an `api` package. I thought we were doing that consistently for JVM SDKs, but maybe not.

npm-js/create-kalix-entity/template/event-sourced-entity/proto/counter_domain.proto
```diff
@@ -4,15 +4,6 @@ package com.example.domain;
 
 import "kalix/annotations.proto";
```
pvlugter, line 5: Should be possible to remove the kalix annotations import from the domain files, no longer used.

samples/ts/ts-eventsourced-shopping-cart/proto/shoppingcart_api.proto
```diff
@@ -22,6 +22,8 @@ import "google/api/annotations.proto";
 import "google/protobuf/empty.proto";
 import "kalix/annotations.proto";
 
+
+
```
pvlugter, line 26: unnecessary blank lines?

pvlugter (commented):
LGTM

samples/ts/ts-eventsourced-shopping-cart/proto/shoppingcart_api.proto
franciscolopezsancho: done

npm-js/create-kalix-entity/template/event-sourced-entity/proto/counter_domain.proto
franciscolopezsancho: done

PR #430, opened 2022-11-10
bump(codegen): Scala 2.13.10, sbt plugin and library updates

ennru:


pvlugter:
Looks like the native image build is failing. Will probably need an updated config for the dependency changes. I think there's a script or something to run to generate new native image config.

johanandren:
Bumped 2.13.10 in #432 but didn't touch any plugins and libraries

PR #432, opened 2022-11-14
chore: upgrade scala version to 2.13.10

efgpinto:
FOSSA is complaining about this so I thought we could make it happier.


efgpinto:
I suspected there was probably a reason for this not be have been done previously. At least I know now: `sbt.librarymanagement.ResolveException: Error downloading org.wartremover:wartremover_2.13.9:2.4.13`

johanandren:
If we bump it we should make it 2.13.10 (for the serialization bugfix) but I don't think it is very important since it is only about codegen and not anything actually running in services.

johanandren:
For 2.13.10 wartremover 2.4.21 should work (but not quite sure why we use it at all, I think scalafmt is enough, so maybe we could remove it from the codegen project)

johanandren (approved):

PR #434, opened 2022-11-14
fix: avoid failing silently when publishing codegen

efgpinto:
Fix #433 

I could not test this locally to confirm which curl version is included but upgraded to latest ubuntu version which is very likely to contain the required curl version `7.76.0`.

I know we are already using this ubuntu version in other places. Let me know if you think this is a bad idea.

PR #436, opened 2022-11-14
chore: bump samples to 1.0.6

efgpinto:
References https://github.com/lightbend/kalix-javascript-sdk/issues/435

Includes also the bump of macos version used in circleci for publishing the native image. Required after curl new parameter added in #434

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -956,7 +956,7 @@ jobs:
 
   publish_native_macos:
     macos:
-      xcode: 12.5.1
+      xcode: 13.4.1
     description: "Build Native image on macOS"
     steps:
       - checkout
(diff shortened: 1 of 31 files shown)


pvlugter:
Big diffs on the package lock files, as it's also bumping to `"lockfileVersion": 3` from version 1. Things here have stayed on older versions of Node and npm. May be useful to continue for now, until these are bumped everywhere. An `nvm use 14` should be on the old versions (unless npm was upgraded manually I guess).

efgpinto:
> Big diffs on the package lock files, as it's also bumping to `"lockfileVersion": 3` from version 1. Things here have stayed on older versions of Node and npm. May be useful to continue for now, until these are bumped everywhere. An `nvm use 14` should be on the old versions (unless npm was upgraded manually I guess).

Thanks Peter, good to know. Updated now using node 14. lockfileVersion is back to 1.

johanandren (approved):

PR #437, opened 2022-11-17
chore: Bump proxy to 1.0.28

johanandren:
https://github.com/lightbend/kalix-proxy/issues/1727

PR #438, opened 2022-12-13
fix: allow updated type dependency for node

pvlugter:
Looking to fix the typescript issue as seen in #437 by allowing an updated `@types/node` dependency.


pvlugter:
Probably due for a round of dependency updates here, but will just do this simple one for now.

CI is happy, so I'll merge to unblock other PRs.

PR #444, opened 2022-12-21
ci: add script to bump proxy version

efgpinto:
Adding this script so we can use it later in automating the proxy release a bit more. For now, it can be used manually anyway.

References https://github.com/lightbend/kalix-proxy/issues/967


octonato (approved):
LGTM

PR #453, opened 2023-02-08
fix: Catch and continue discovery if component.preStart fails

johanandren:
For example if it is not defined, or overloaded with broken logic.

References lightbend/kalix-proxy#1570


efgpinto (approved):
LGTM

Issue #389, opened 2022-07-06
Report unique SDK names for JavaScript and TypeScript


johanandren:
At discovery the SDK tells the proxy what SDK it is, it currently reports the same "kalix-javascript-sdk" for both JavaScript and TypeScript. We should figure out which and report "kalix-javascript-sdk", "kalix-typescript-sdk" instead if possible.

Note that we need to make sure that the proxy SDK release version hinting knows about the names so we get the correct "there's a new SDK version out"

PR #454, opened 2023-02-08
chore: Report ts and js as separate SDKs #389

johanandren:
References #389

Depends on upstream support for the typescript name in the proxy that must be in production before this can be merged/released.


efgpinto (approved):
LGTM

Issue #455, opened 2023-02-09
Un-filter eslint errors


johanandren:
In #452 I downgraded a bunch of eslint errors to warnings to get the upgrade to pass, we should go through and see if we can fix instead of filter each rule.

_Originally posted by @efgpinto in https://github.com/lightbend/kalix-javascript-sdk/pull/452#discussion_r1100744470_

PR #457, opened 2023-02-09
Es-lint errors followup

johanandren:
References #455


johanandren:
I wonder if we should make it fail on errors in ci like we do for Scala projects?

efgpinto (approved):
LGTM

PR #476, opened 2023-03-27
chore: configure doc linting in specific dirs

beritou:


docs/src/modules/javascript/pages/call-another-service.adoc
```diff
@@ -6,7 +6,7 @@ In some cases it is useful to call a component in another service, for example i
 
 Calling other Kalix services in the same project from an Action is done by invoking them over gRPC much like how an external client would. The service is however identified only by the name it has been deployed as, Kalix takes care of routing requests to the service and keeping the data safe by encrypting the connection for us.
 
-In this sample we will show an action that does two sequential calls to the xref:value-entity.adoc[Value Entity Counter] service, deployed with the service name "counter".
+In this sample we will show an action that does two sequential calls to the xref:value-entity.adoc[Value Entity Counter] service, deployed with the service name "counter."
```
marcoderama, line 9: I know this is correct English style, but it always troubles me when we're quoting some important literal like this.

If that correction is unavoidable on the end of a sentence, perhaps we could go with something like
> ... deployed with the "counter" service name.

Just an idea...  Not critical.

marcoderama (approved):
approved, but couldn't help adding a small comment...  Feel free to ignore.

docs/src/modules/javascript/pages/call-another-service.adoc
efgpinto: I think another option in these cases is to use "`counter`."

PR #482, opened 2023-05-16
fix: hide todos from published docs

octonato:
Same as https://github.com/lightbend/kalix-jvm-sdk/pull/1625

### DIFF
--- a/docs/dev/antora.yml
+++ b/docs/dev/antora.yml
@@ -17,8 +17,6 @@ runtime:
 
 asciidoc:
   attributes:
-    review: ''
-    todo: ''
     page-pagination: ''
     page-toctitle: 'On This Page'
 


ennru (approved):
LGTM.

docs/dev/antora.yml
```diff
@@ -17,8 +17,6 @@ runtime:
 
 asciidoc:
   attributes:
-    review: ''
-    todo: ''
```
pvlugter, line 21: Isn't this just the local dev view? Not included in published docs?

octonato: Hmm, correct. I missed the fact that it was in `dev` path. 

I will check it once again. There must be some other properties or we are using dev to publish the docs, because we do have the TODOs showing up in docs.kalix.io.

octonato: And indeed, we want to see the TODOs when building locally and my change remove it completely.

PR #483, opened 2023-05-16
chore: build docs with latest image

octonato:
Relates to https://github.com/lightbend/kalix-docs/pull/1773


octonato:
I'm giving up on this one. Antora image is not built for arm64 anyway and we are having other issues now related with xref-validator. Not worth spent time on this for now.

PR #484, opened 2023-05-17
fix: re-enable todos in dev

octonato:
Reverts #482 

These settings are used when building the docs locally in dev mode. Something else is enabling this in prod.


ennru (approved):
Sorry for the confusion.

PR #490, opened 2023-05-26
chore: bump to node v18.16

octonato:
As suggested here https://github.com/lightbend/kalix-javascript-sdk/pull/459/files#r1177168357


testkit/package.json
```diff
@@ -31,13 +31,13 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "peerDependencies": {
-    "@kalix-io/kalix-javascript-sdk": "1.x"
+    "@kalix-io/kalix-javascript-sdk": "file:../sdk/kalix-io-kalix-javascript-sdk-0.0.0.tgz"
```
johanandren, line 34: This looks wrong?

octonato: Absolutely. No idea how that happened though. Especially because it's on the package.json file. 

I will check/revert/whatever needed

octonato: I was looking around to understand how this got updated and it comes from `./bin/create-local-dependencies.sh` script. This scripts seems to force the update in `testkit/package.json`. Not sure if this is intentional or desirable. 

I will revert back to `1.x`, but looking for guidance from someone with knowledge in the JS space.

octonato: I remember Andrea saying something about the semantics of a lib file with `0.0.0` version. But I can recall the details. There was some trick/hack around it, but I'm not sure if that's what we need here. 

Maybe @pvlugter or @katsutoxin can shed a light?

katsutoxin: The file `create-local-dependencies.sh` is created by @franciscolopezsancho , maybe he has better idea?  I "guess" that file is only for local development, maybe we should not check it in. (it is my personal guess)

pvlugter: Yes, will just be for testing locally. Means that it's best to make changes and commit them, then run tests and discard the changes.

octonato:
It seems that I managed to make it green! 🎉

johanandren (approved):

testkit/package.json
franciscolopezsancho: wow, totally missed this comment. Sorry @katsutoxin. Yes, that file is only for local development.

PR #525, opened 2023-12-13
Switch to "Kalix Runtime"

ennru:
References
- https://github.com/lightbend/kalix/issues/9895
- https://github.com/lightbend/kalix-runtime/pull/2173

Supersedes
- #522

### DIFF
--- a/sdk/bin/prepare.sh
+++ b/sdk/bin/prepare.sh
@@ -21,12 +21,12 @@ if [ -d "./proto" ] ; then
   echo "Protocol already built ('npm run clean' first to fetch and compile again)"
   echo
 else
-  if [ -n "$PROXY_SNAPSHOT_DIRECTORY" ]; then
-     # Use local proxy and sdk sources, useful for development, point PROXY_SNAPSHOTS_DIRECTORY to the local
-     # proxy project source directory
-     echo "Using snapshot of proxy and sdk protocols from '$PROXY_SNAPSHOT_DIRECTORY'"
-     cp -rf $PROXY_SNAPSHOT_DIRECTORY/protocols/proxy/src/main/protobuf/* ./proto/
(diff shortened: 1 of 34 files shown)


sdk/bin/prepare.sh
```diff
@@ -21,12 +21,12 @@ if [ -d "./proto" ] ; then
...
-     # proxy project source directory
-     echo "Using snapshot of proxy and sdk protocols from '$PROXY_SNAPSHOT_DIRECTORY'"
-     cp -rf $PROXY_SNAPSHOT_DIRECTORY/protocols/proxy/src/main/protobuf/* ./proto/
-     cp -rf $PROXY_SNAPSHOT_DIRECTORY/protocols/sdk/src/main/protobuf/* ./proto/
+  if [ -n "$RUNTIME_SNAPSHOT_DIRECTORY" ]; then
+     # Use local runtime and sdk sources, useful for development, point RUNTIME_SNAPSHOT_DIRECTORY to the local
+     # runtime project source directory
+     echo "Using snapshot of proxy and sdk protocols from '$RUNTIME_SNAPSHOT_DIRECTORY'"
+     cp -rf $RUNTIME_SNAPSHOT_DIRECTORY/protocols/proxy/src/main/protobuf/* ./proto/
+     cp -rf $RUNTIME_SNAPSHOT_DIRECTORY/protocols/sdk/src/main/protobuf/* ./proto/
```
efgpinto, line 29: Where do we set this ENV VAR `RUNTIME_SNAPSHOT_DIRECTORY` ? I could not find it.. jsut making sure we update that piece as well

efgpinto (approved):
LGTM with a question.

sdk/bin/prepare.sh
ennru: My understanding is that this done manually for local development when you want to develop against an unreleased Runtime version with sources on your machine.

efgpinto: ah right, makes sense 👍

PR #530, opened 2023-12-21
ci: update env var for runtime version

efgpinto:
Refs:
- https://github.com/lightbend/kalix/issues/9895


.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
ennru, line 1: I guess this was not intended.

bin/update-proxy-versions.sh
```diff
@@ -1,32 +1,32 @@
 #!/usr/bin/env bash
 
 # USAGE:
-# > PROXY_VERSION=1.0.31 ./update-proxy-versions.sh
```
ennru, line 4: Should we rename the script, as well? (Preferably less broken than I did in the JVM SDK.)

bin/update-proxy-versions.sh
```diff
@@ -1,32 +1,32 @@
...
+# > RUNTIME_VERSION=1.0.31 ./update-proxy-versions.sh
 
 # this script is meant to be used after a new Proxy version is out
 # to facilitate the update of all the places where we usually depend on the latest version
 
 # provide the new proxy version you want the project to be updated to
```
ennru, line 9:
```suggestion
# provide the new Kalix Runtime version you want the project to be updated to
```

ennru (commented):
I'd even here generally prefer the `${...}` syntax.
LGTM.

bin/update-proxy-versions.sh
efgpinto: Yes, lets do that. https://github.com/lightbend/kalix-runtime/pull/2271

ennru (approved):

efgpinto:
We are getting `429 Too Many Requests` for `https://hub.docker.com/_/node` .. I will retry later.

PR #544, opened 2024-01-25
Bump: runtime 1.1.30

johanandren:
Replaces #543

### DIFF
--- a/testkit/src/integration-testkit.ts
+++ b/testkit/src/integration-testkit.ts
@@ -175,7 +175,7 @@ export class IntegrationTestkit {
         'VERSION_CHECK_ON_STARTUP',
         process.env.VERSION_CHECK_ON_STARTUP || 'true',
       )
-      .withWaitStrategy(Wait.forLogMessage('Starting Kalix '))
+      .withWaitStrategy(Wait.forLogMessage('Connecting to '))
       .start();
     this.proxyPort = proxyContainer.getMappedPort(9000);
     // sdk needs to know how to call itself for cross component calls,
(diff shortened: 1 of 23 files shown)


johanandren:
Not sure what was going on, couldn't push to the original auto-bump PR, so created my own branch, but then that commit ended up in the original bump PR anyway somehow. 😕

testkit/src/integration-testkit.ts
```diff
@@ -175,7 +175,7 @@ export class IntegrationTestkit {
         'VERSION_CHECK_ON_STARTUP',
         process.env.VERSION_CHECK_ON_STARTUP || 'true',
       )
-      .withWaitStrategy(Wait.forLogMessage('Starting Kalix '))
+      .withWaitStrategy(Wait.forLogMessage('Connecting to '))
```
johanandren, line 178: Seems the order of discovery and logs changed somewhat in the runtime, causing a deadlock because waiting for the `Starting Kalix ` log output, but that was only logged once discovery completes, and discovery would not complete until the test container proxy port is injected a few lines down from here.

johanandren:
Meh, all green but circle did not report back properly.

johanandren:
The same again, all steps of the CI workflow green over at Circle but not reported back here.

testkit/src/integration-testkit.ts
aludwiko: that's odd, the line is still there...

aludwiko: I mean in the runtime code

aludwiko (approved):

